PR #2, opened 2019-10-15
Server

hannesm:
this is an initial server implementation, it has been tested with a local openvpn client (tcp only). what is missing is to actually forward data packets anywhere (currently only icmp echo replies are answered to), address management is slightly weak.

but overall, the diff is reasonably small and it is pretty nice that (even without the explicit goal) we managed to be able to reuse quite a lot of code from the client in the server :)


hannesm:
thanks for your review, sorry for the delay. I addressed your comments in 1a84aad and rebased onto master.

while the server implementation may still not be ready-for-production, I'm keen to get this PR merged (the rebase was pretty labour-intensive and I'd like to avoid yet another rebase).

mirage-server/unikernel.ml
```diff
@@ -0,0 +1,23 @@
...
+
+module Main (R : Mirage_random.C) (M : Mirage_clock.MCLOCK) (P : Mirage_clock.PCLOCK) (T : Mirage_time_lwt.S) (S : Mirage_stack_lwt.V4) (FS: Mirage_kv_lwt.RO) = struct
+
+  module O = Openvpn_mirage.Server(R)(M)(P)(T)(S)
+
+  let read_config data =
+    FS.get data (Mirage_kv.Key.v "openvpn.config") >|= function
+    | Error e -> Rresult.R.error_to_msg ~pp_error:FS.pp_error (Error e)
+    | Ok data ->
+      let string_of_file _ = Error (`Msg "no string_of_file support") in
```
cfcs: Since this is a user-facing error, perhaps something like "The openvpn configuration contains references to external files, which this unikernel cannot read. Please inline those external files into the provided configuration." or something like that. Not blocking for this PR, just something we can consider in the future (since we do have a utility to do this inlining).

PR #6, opened 2019-11-05
adapt to new mirageos interfaces

hannesm:


cfcs:
Is there a PR for `mirage-nat` that we can track to merge this once that is merged?

hannesm:
yes indeed https://github.com/mirage/mirage-nat/pull/34 ... but there also failing CI, I'll look what is the issue there

hannesm:
hmm, interesting -- here CI ran twice (PR and PUSH), where one succeeded and one failed on Qubes -- the failure is very similar to the one in mirage-nat, namely `Unbound value Key_gen.private_ipv4`. Looks like there's somewhere a race condition, unclear to me where.

PR #7, opened 2019-11-23
Re enable app

hannesm:
this is on top of #6 and adapts the API changes of the openvpn interface for the lwt client.

//cc @cfcs


hannesm:
CI is drunk, it needs for the `mirage-nat` unikernel a pin of mirage-nat to hannesm/mirage-nat#easy... will try to work on mirage-nat (+release) tomorrow to unblock CI.

hannesm:
in addition, once this and #1 is in (maybe #2 as well), its likely a good idea to refactor the app/openvpn client and mirage/openvpn_mirage (atm there's quite some code duplication with subtle changes)

PR #8, opened 2019-12-19
changes for mirage-nat 2.0.0

hannesm:
ready to be merged once https://github.com/ocaml/opam-repository/pull/15572 is in

### DIFF
--- a/mirage-nat/unikernel.ml
+++ b/mirage-nat/unikernel.ml
@@ -36,8 +36,15 @@ module Main (R : Mirage_random.S) (M : Mirage_clock.MCLOCK) (P : Mirage_clock.PC
       | Ok ovpn ->
         Logs.info (fun m -> m "tunnel established");
         let output_tunnel packet =
-          O.write ovpn (Cstruct.concat (Nat_packet.to_cstruct packet)) >|= fun res ->
-          if not res then Log.err (fun m -> m "failed to write data via tunnel")
+          match Nat_packet.to_cstruct ~mtu:(O.mtu ovpn) packet with
+          | Ok pkts ->
+            Lwt_list.fold_left_s
+              (fun r p -> if r then O.write ovpn p else Lwt.return r)
+              true pkts >|= fun res ->
+            if not res then Log.err (fun m -> m "failed to write data via tunnel")
+          | Error e ->
+            Log.err (fun m -> m "NAT to_cstruct failed %a" Nat_packet.pp_error e);
+            Lwt.return_unit
         and output_private packet =
           let dst = match packet with `IPv4 (p, _) -> p.Ipv4_packet.dst in
           Private_routing.destination_mac private_ip_net None arp dst >>= function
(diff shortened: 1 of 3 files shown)


mirage-nat/unikernel.ml
```diff
@@ -46,16 +53,25 @@ module Main (R : Mirage_random.S) (M : Mirage_clock.MCLOCK) (P : Mirage_clock.PC
                         (match e with `Local -> "local" | `Gateway -> "gateway"));
             Lwt.return_unit
           | Ok dst ->
+            let more = ref [] in
             E.write eth dst `IPv4 (fun b ->
```
cfcs: what is `b` here? could we name it `buf` ?

mirage-nat/unikernel.ml
```diff
@@ -46,16 +53,25 @@ module Main (R : Mirage_random.S) (M : Mirage_clock.MCLOCK) (P : Mirage_clock.PC
...
                 match Nat_packet.into_cstruct packet b with
-                | Ok n -> n
+                | Ok (n, adds) -> more := adds ; n
```
cfcs: is `n` guaranteed to be <= `sizeof b` ? should we `assert` ?

mirage-nat/unikernel.ml
```diff
@@ -46,16 +53,25 @@ module Main (R : Mirage_random.S) (M : Mirage_clock.MCLOCK) (P : Mirage_clock.PC
...
                 | Error e ->
                   Log.err (fun m -> m "error %a while Nat_packet.into_cstruct"
                               Nat_packet.pp_error e);
-                  0) >|= function
```
cfcs: will `0` trigger the `Error` case below? (if not, should we record an error condition? if yes, should we leave a comment for the next reviewer that this is how it works?)

mirage-nat/unikernel.ml
```diff
@@ -69,25 +85,27 @@ module Main (R : Mirage_random.S) (M : Mirage_clock.MCLOCK) (P : Mirage_clock.PC
             | Ok packet -> output_tunnel packet
             | Error `Untranslated -> add_rule table packet
             | Error `TTL_exceeded ->
+              (* TODO should report ICMP error message to src *)
               Log.warn (fun f -> f "TTL exceeded"); Lwt.return_unit
```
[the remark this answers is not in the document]
hannesm: but it is not necessarily ICMP... the on the private interface received packet has a TTL of 1, i.e. it should not be routed further. this may be any IP packet...

mirage-nat/unikernel.ml
```diff
@@ -36,8 +36,15 @@ module Main (R : Mirage_random.S) (M : Mirage_clock.MCLOCK) (P : Mirage_clock.PC
...
         Logs.info (fun m -> m "tunnel established");
         let output_tunnel packet =
-          O.write ovpn (Cstruct.concat (Nat_packet.to_cstruct packet)) >|= fun res ->
-          if not res then Log.err (fun m -> m "failed to write data via tunnel")
+          match Nat_packet.to_cstruct ~mtu:(O.mtu ovpn) packet with
+          | Ok pkts ->
+            Lwt_list.fold_left_s
+              (fun r p -> if r then O.write ovpn p else Lwt.return r)
+              true pkts >|= fun res ->
+            if not res then Log.err (fun m -> m "failed to write data via tunnel")
```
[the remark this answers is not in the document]
hannesm: that's information available to the `O.write` -- which in turn calls transmit which emits such log messages :) here on the unikernel layer, there's no information whether UDP or TCP was used for the tunnel

mirage-nat/unikernel.ml
```diff
@@ -46,16 +53,25 @@ module Main (R : Mirage_random.S) (M : Mirage_clock.MCLOCK) (P : Mirage_clock.PC
...
                 match Nat_packet.into_cstruct packet b with
-                | Ok n -> n
+                | Ok (n, adds) -> more := adds ; n
                 | Error e ->
                   Log.err (fun m -> m "error %a while Nat_packet.into_cstruct"
                               Nat_packet.pp_error e);
-                  0) >|= function
-            | Ok () -> ()
+                  0) >>= function
+            | Ok () ->
```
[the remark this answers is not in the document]
hannesm: done

mirage-nat/unikernel.ml
hannesm: it is (due to me knowing the mirage_nat implementation) ;)

mirage-nat/unikernel.ml
hannesm: the contract^W signature of [`E.write` is `Cstruct.t -> int` for the `fill` function](https://github.com/mirage/mirage-protocols/blob/v4.0.1/src/mirage_protocols.mli#L32-L37), i.e. there's no way to error when calling `write` in the fill function. The return value (`int`) is used by the ethernet layer to determine how many bytes to send, if this case here is visited, an empty ethernet frame (i.e. only the header) will be transmitted (not actually sure whether it will be transmitted). this no-error in fill may actually be a design mistake we should re-think.

cfcs: Yeah, it seems like something that could fail.

PR #12, opened 2020-02-03
More configuration options

hannesm:
the first commit is #10 (which imho is fine to merge as is). the topic of this PR is the second commit, adding some more configuration options:
- ~~`rport p` (replacing all _default_ port values (=1194) in remotes by p)~~
- typo fix
- dev_type (parses, part of the Config GADT, not used by logic)
- ignore `script_security`

the motivation for this change is actual configuration files I use and have on my disk //cc @cfcs

### DIFF
--- a/src/config.ml
+++ b/src/config.ml
@@ -60,6 +60,7 @@ module Conf_map = struct
     | Connect_retry_max : [`Unlimited | `Times of int] k
     | Connect_timeout : int k
     | Dev      : [`Null | `Tun of int option | `Tap of int option] k
+    | Dev_type : [ `Tun | `Tap ] k
     | Dhcp_disable_nbt: flag k
     | Dhcp_dns: Ipaddr.t list k
     | Dhcp_ntp: Ipaddr.t list k
@@ -179,7 +180,7 @@ module Conf_map = struct
         (X509.Private_key.encode_pem key |> Cstruct.to_string)
     in
     match k,v with
-    | Auth_nocache, () -> p() "auto-nocache"
+    | Auth_nocache, () -> p() "auth-nocache"
     | Auth_retry, `None -> p() "auth-retry none"
     | Auth_retry, `Nointeract -> p() "auth-retry nointeract"
     | Auth_retry, `Interact -> p() "auth-retry interact"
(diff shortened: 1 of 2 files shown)


src/config.ml
```diff
@@ -1108,7 +1122,7 @@ let valid_server_options ~client:_ _server_t =
 
 let parse_next (effect:parser_effect) initial_state : (parser_state, 'err) result =
   let open Rresult in
-  let rec loop (acc:Conf_map.t) : line list -> (parser_state,'b) result =
+  let rec loop (acc:Conf_map.t) rport : line list -> (parser_state,'b) result =
```
cfcs: I'm a little bit skeptical of adding a parameter for `rport` to this function.
I think a better approach may be to make the lexer for `remote` register a polymorphic variant `Default_port`, and rewriting in the end (before return to the consumer) to be able to correctly parse:
```
remote foo.example 1194
rport 1234
remote bar.example
```
which should result in
```
Remote foo.example 1194
Remote bar.example 1234
```
I think this implementation (unless I'm reading it wrong) would result in
```
Remote foo.example 1234
Remote bar.example 1234
```

(idk, maybe that would be a good test case to add). In any case I'm okay with merging this PR as-is.

hannesm:
force-pushed without the rport changes; they seem to need some more work. from experimentation with C's openvpn:
```
remote 10.0.42.5
remote 10.0.42.3 1194
rport 1234
remote 10.0.42.4
```

this leads to three connection profiles:
10.0.42.5 1234
10.0.42.3 1194
10.0.42.4 1234

--> the line of rport is not relevant, it overwrites all non-specified ports in remote entries. but you are right, if 1194 is specified explicitly, rport should leave this alone.

src/config.ml
hannesm: > I think a better approach may be to make the lexer for remote register a polymorphic variant `Default_port`, and rewriting in the end (before return to the consumer)

I'd be in favour of keeping the key-value of `Remote` as is in our config GADT. I thought loop is exactly the right place to transform other line directives to configuration map elements. I don't quire understand what "in the end" in your sentence should mean.

PR #14, opened 2020-02-04
config: add test suite and triggers for two bugs

cfcs:


test/config_afl.ml
```diff
@@ -0,0 +1,11 @@
+let () =
+  let fd = Unix.openfile Sys.argv.(1) Unix.[O_RDONLY] 0 in
+  let {Unix.st_size = len ; _ } = Unix.fstat fd in
+  let input =
+    let b = Bytes.create len in (* "best effort": *)
+    Bytes.sub_string b 0 (Unix.read fd b 0 len) in
```
hannesm: putting the `in` on the next line with indentation = 2 spaces makes it much easier to read imho...

test/config_tests.ml
```diff
@@ -0,0 +1,124 @@
+let pmsg =
+  Alcotest.testable (fun ppf (`Msg s) -> Fmt.pf ppf "Error @[<v>(%s)@]" s)
+    (=)
```
hannesm:
```suggestion
    (fun (`Msg a) (`Msg b) -> String.equal a b)
```

let's avoid polymorphic equality when we can :)

test/config_tests.ml
```diff
@@ -0,0 +1,124 @@
...
+
+let result_p_err (type x) ok =
+  (* like Alcotest.result, but pretty-prints the Error `Msg and
+     fails the test when [Error _] is encountered.
+     On success, [ok]'s [pp] and [equal]
+     function are used to pretty-print and determine test result. *)
```
hannesm: hmm, why not use `result <ok> p_msg` (with `p_msg` defined above) at call-sites?

test/config_tests.ml
```diff
@@ -0,0 +1,124 @@
...
+  let basic =
+    {|tls-client
+    auth-user-pass [inline]
+    <auth-user-pass>
+testuser
+testpass
+</auth-user-pass>
+remote 10.0.0.1|} in
+  Alcotest.(check (result pass pmsg)) "basic conf works"
+    (Ok Openvpn.Config.empty)
```
hannesm: I don't understand why `Openvpn.Config.empty` is the result of this config test -- I'd expect there being at least remote and the auth-user-pass keys and values being present.

test/config_tests.ml
```diff
@@ -0,0 +1,124 @@
...
+     It should also be tested if the upstream version strips prefixed/trailing
+     whitespace from the user/pass/end-block lines. TODO.
+  *)
+  let common =
+    "client\n"
+    ^ "remote 127.0.0.1\n"
+    ^ "auth-user-pass [inline]\n"
+    ^ "<auth-user-pass>\n"
+    ^ "testuser\n"
+    ^ "testpass\n" in
```
hannesm: I find the multi-line string literal syntax (`{| ... |}`) easier to read than these concatenated strings..

test/config_tests.ml
```diff
@@ -0,0 +1,124 @@
...
+    <auth-user-pass>
+testuser
+testpass
+</auth-user-pass>
+
+    remote 10.0.42.5 1234
+    remote 10.0.42.3 1194
+    rport 1234
+    remote 10.0.42.4 1234
+|} |> parse_client ~string_of_file:(fun _ -> Rresult.R.error_msg "")
```
hannesm: you could use the `parse_noextern` from above here

test/config_tests.ml
```diff
@@ -0,0 +1,124 @@
...
+  () (* TODO check that the ports and remotes also match the written *)
+
+let crowbar_fuzz_config () =
+  AflPersistent.run (fun () ->
+  Crowbar.add_test ~name:"Fuzzing doesn't crash Config.parse_client"
+    [Crowbar.bytes] (fun s ->
+        try Crowbar.check (ignore @@ parse_noextern s ; true)
+        with _ -> Crowbar.bad_test ()
+      )
+    )
```
hannesm: I'd recommend to put the unit/regression tests of configuration in a separate binary than the crowbar fuzzing (due to dependencies -- it's nice to run the unit tests on each commit, while fuzzing could be only on drastic changes to the configuration parser).

The fuzzing part atm checks that `parse_noextern` does not crash (IIUC) -- we may as well generate `Conf_map.t` values, serialise (generating a string) and parse that (getting another Conf_map.t) and check that (a) the parse succeeds, (b) the two Conf_map.t are equal. this would catch things like typos in serialiser / parser... (maybe out of scope for this PR)

hannesm:
thanks for this PR, i commented with some suggestions

test/config_tests.ml
cfcs: Because I was being lazy and put `Alcotest.pass` on the `ok`-side of `Alcotest.result`, so I just needed "a value."

I think a cleaner design would be to have a global `Config.t` that represented this minimal config, since it looks like we end up repeating it quite a lot since all tests need `client`, `remote`, and a way to authenticate atm. I will move things around a bit.

test/config_tests.ml
cfcs: I had some kind of horrible hack going on locally where I kept track of which one was expected/actual and colored them differently because I found it quite hard to visually tell the difference between two long config files, but I think you're right.

test/config_tests.ml
cfcs: implemented that now by copy-pasting over. I didn't like the idea of exposing it in `openvpn.mli`, since it lacks some keys (`remote`, auth, ...) and thus doesn't pass validation without additions.

On that note, wonder if we should expose the `is_valid_client_config` function.

test/config_tests.ml
cfcs: I had problems with auto-indentation there since the semantics are different; leading whitespace in `{| |}` is included in the string, and this test explicitly deals with whitespace, so I wanted to be sure of what ended up in this test even after eventual reformatting. Generally I agree :)

test/config_tests.ml
cfcs: Crowbar runs pretty fast for me, but if you're worried about the dependency, sure. I already have a separate file for `afl-fuzz` that I should also commit.

Good test to have, but yes, I think in a different PR, maybe using the Crowbar+`deriving` tooling.

hannesm:
this PR now includes commits from #13 and #12 -- could you rebase onto master without these commits and force-push, please? This then allows to view the diff without the noise. Thanks.

cfcs:
Sorry, not sure what happened there. Should be fixed now.

test/config_tests.ml
hannesm: ok, I agree that crowbar + deriving would be nice there. can we move the afl-persistent / crowbar dependent test to `config_afl.ml`? I usually develop in switches that are not based on afl, and as mentioned above would appreciate to run the unit tests nevertheless during development.

cfcs: Yes, will do that :)

PR #16, opened 2020-02-06
Config: now that dev_type is supported, use this in client_generate_connect_options as well

hannesm:
missing from #12

### DIFF
--- a/src/config.ml
+++ b/src/config.ml
@@ -1420,9 +1420,9 @@ let client_generate_connect_options t =
       | _ -> false) t in
   let serialized =
     (Fmt.strf "V4,tls-client,%s%a"
-       (match Conf_map.find Dev t with
-        | Some `Tun _ -> "dev-type tun,"
-        | Some `Tap _ -> "dev-type tap,"
+       (match Conf_map.find Dev t, Conf_map.find Dev_type t with
+        | Some `Tun _, _ | None, Some `Tun -> "dev-type tun,"
+        | Some `Tap _, _ | None, Some `Tap -> "dev-type tap,"
         | _ -> "")
        (Conf_map.pp_with_sep ~sep:(Fmt.unit ",")) excerpt) in
   Logs.warn (fun m -> m "serialized connect options, probably incorrect: %S"


src/config.ml
```diff
@@ -1420,9 +1420,9 @@ let client_generate_connect_options t =
       | _ -> false) t in
   let serialized =
     (Fmt.strf "V4,tls-client,%s%a"
-       (match Conf_map.find Dev t with
-        | Some `Tun _ -> "dev-type tun,"
-        | Some `Tap _ -> "dev-type tap,"
+       (match Conf_map.find Dev t, Conf_map.find Dev_type t with
+        | Some `Tun _, _ | None, Some `Tun -> "dev-type tun,"
+        | Some `Tap _, _ | None, Some `Tap -> "dev-type tap,"
```
cfcs: so if I have
```
dev tun0
dev-type tap
```
It will use `tun` - ? should we fail in that case?

hannesm: hmm, interesting... maybe there should be only ``Dev : [`Null | `Tun of int option | `Tap of int option] k`` in the Conf_map (and `dev-type tun` will become ``Dev (`Tun None)``)?

I am not sure why there's `` `Null `` in the variant. Is a `dev tun` (i.e. without a number) different from `dev-type tun` at all?

hannesm: (and then in loop where we insert stuff into the Conf_map, we should avoid `dev-type` overriding `dev`, and fail on contradictionary `dev-type tap` and `dev tun25`)...

cfcs: My reading of `man openvpn` leaves me with the understanding that specifying `dev-type` without `dev` is illegal, and that `dev-type` is only useful in a config like this:
```
dev-type tun
dev openvpnif # tun interface with custom name
```

I have no idea whatsoever what `dev null` is supposed to do. Looking at this post https://forums.openvpn.net/viewtopic.php?t=14057 it looks like it was broken at some point (is it still?), so maybe we should just remove that. It's only in the parser because it was listed in the manpage.

cfcs: also not clear to me with `dev customname` is valid; the docs for `--dev` suggest not, but the docs for `dev-type` suggest that `dev` should just take a string, where `tun[0-9]*` + `tap[0-9]*` have special semantics.

cfcs: I would kind of lean towards a conservative approach of:
```ocaml
Dev : [`Tun of int option | `Tap of int option] k
```

Only accepting things like this:
```
dev tap
dev tun
dev tap0
dev tun0
dev tap123
dev tun123
```

With `dev-type` being an optional constraint on `dev`, so they must match.

What do youthink about that approach?

cfcs: (Thus not exposing `Dev_type` in the `Config_map.t`, and only keeping it around as a polymorphic variant)

hannesm:
to sum up, as discussed: Conf_map gets two keys `Dev : string k` and ``Dev_type : [ `Tun | `Tap ] k``.

a configuration entry `dev tun100` should insert both `Dev "tun100"` and ``Dev_type `Tun`` into the configuration map (unless `dev-type` is already present). Whether to fail on
```
dev-type tap
dev tun123
```
is not clear, if not the Conf_map should contain `Dev tun123` and ``Dev_type `Tap`` (i.e. do not overwrite with `` `Tun``).

hannesm:
superseeded by #17

PR #17, opened 2020-02-06
config: better handling of [dev] and [dev-type], and tests!

cfcs:
Supercedes https://github.com/roburio/openvpn/pull/16

### DIFF
--- a/app/openvpn_client_lwt.ml
+++ b/app/openvpn_client_lwt.ml
@@ -6,10 +6,12 @@ let open_tun config {Openvpn.ip ; gateway ; _ }
      the TUN interface *)
   let open Lwt_result.Infix in
   begin match Openvpn.Config.find Dev config with
-    | None | Some `Tun None -> Ok None
-    | Some `Tun (Some n) -> Ok (Some ("tun" ^ string_of_int n))
-    | Some `Null -> Error (`Msg "TODO what is this")
-    | Some `Tap _ -> Error (`Msg "using a TAP interface is not supported")
+    | None | Some (`Tun, None) -> Ok None
+    | Some (`Tun, Some name) -> Ok (Some name)
+    | Some (`Tap, name) ->
+      Rresult.R.error_msgf
+        "using a TAP interface (for %S) is not supported"
+        (match name with Some n -> n | None -> "dynamic device")
   end |> Lwt_result.lift >>= fun devname ->
   try begin
     let fd , dev = Tuntap.opentun ?devname () in
--- a/src/config.ml
+++ b/src/config.ml
@@ -61,8 +61,7 @@ module Conf_map = struct
     | Connect_retry : (int * int) k
     | Connect_retry_max : [`Unlimited | `Times of int] k
     | Connect_timeout : int k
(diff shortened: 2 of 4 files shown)


app/openvpn_client_lwt.ml
```diff
@@ -6,10 +6,12 @@ let open_tun config {Openvpn.ip ; gateway ; _ }
...
-    | None | Some `Tun None -> Ok None
-    | Some `Tun (Some n) -> Ok (Some ("tun" ^ string_of_int n))
-    | Some `Null -> Error (`Msg "TODO what is this")
-    | Some `Tap _ -> Error (`Msg "using a TAP interface is not supported")
+    | None | Some (`Tun, None) -> Ok None
+    | Some (`Tun, Some name) -> Ok (Some name)
+    | Some (`Tap, name) ->
+      Rresult.R.error_msgf
+        "using a TAP interface (for %S) is not supported"
+        (match name with Some n -> n | None -> "dynamic device")
```
hannesm: you could use `Fmt.(option ~none:(unit "dynamic device") string) name` here

src/config.ml
```diff
@@ -1297,6 +1318,38 @@ let parse_next (effect:parser_effect) initial_state : (parser_state, 'err) resul
...
           (* TODO consult `Proto_force *)
           retb (B(Remote, [host, port, proto]))
+        | (`Dev _ | `Dev_type _) as current ->
+          (* there must be a corresponding `Dev or `Dev_type in [tl]*)
+          let typs, tl = List.partition (function
+              | `Dev_type _ -> true
+              | _ -> false) (current::tl) in
+          let names, tl = List.partition (function
+              | `Dev _ -> true
+              | _ -> false) tl in
```
hannesm: the two `List.partition` increase the time complexity of `loop` (instead of O(n), it is now O(3 * n)). fine with me, an alternative could be to pass dev/dev-type as argument to loop (similar to acc).

src/config.ml
```diff
@@ -1297,6 +1318,38 @@ let parse_next (effect:parser_effect) initial_state : (parser_state, 'err) resul
...
+          begin match typs, names with
+            | (`Dev_type typ)::_, [`Dev name]
```
hannesm: I'd be more strict, and disallow multiple `dev-type` lines -- i.e. match here on `` | [`Dev_type typ],  [`Dev name]``

app/openvpn_client_lwt.ml
cfcs: I guess, but I can't remember the `Fmt` converter for `%S` (which quotes/escapes special chars), and I think this is easier to read?

src/config.ml
cfcs: Originally I always emitted `Dev` and `Dev_type` polyvariants here, but after making `tun0` result in a `Dev` in the gmap directly, this is indeed true - thanks for spotting that! I've revised to do as you suggested.

src/config.ml
cfcs: done as you suggested; it's also much easier to read :)

hannesm:
thanks, that looks good to me

PR #19, opened 2020-02-20
Config: rport and auth_user_pass fixes

cfcs:
This extends the test suite, and makes it pass :)

### DIFF
--- a/src/openvpn.mli
+++ b/src/openvpn.mli
@@ -118,8 +118,18 @@ module Config : sig
                   * [`Udp | `Tcp of [`Server | `Client] option]) k
     (** TODO should Proto be bound to a remote? *)
 
-    | Remote : ([ `Domain of [ `host ] Domain_name.t * [`Ipv4 | `Ipv6 | `Any]
-                | `Ip of Ipaddr.t] * int * [`Udp | `Tcp]) list k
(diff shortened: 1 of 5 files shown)


hannesm:
rebased and force pushed to have a CI run

src/openvpn.mli
```diff
@@ -118,8 +118,18 @@ module Config : sig
...
+                                * [`Ipv4 | `Ipv6 | `Any]
+                   | `Ip of Ipaddr.t]
+                   * [`Port of int | `Default_rport]
+                   * [`Udp | `Tcp]) list
+                 * [`Rport of int] ) k
+    (** [Remote (peers, rport)] specifies the list of peers to
+        connect to.
+        [rport] is the port number to use when a peer has [`Default_rport].
+        Each peer consists of the tuple [address,port,protocol].
+    *)
```
hannesm: hmm, this change makes it much harder to use the parsed configuration. now, the client need to do multiple lookups to figure out the remote. I'd be in favour, similar to how I attempted to deal with `rport`, to handle it internally in the configuration parser, and don't expose it to a user. as user of the configuration I'm interested in "hostname / ip", port, protocol.

hannesm:
thanks for your update (I just noticed that there was a push here), this now is fine with me, and should be merged :D

PR #31, opened 2020-03-28
Allow client configs that have client cert+key without user/pass (#24)

### LINKED ISSUE #24: config-parser refuses config without user/password
openvpn-config-parser client1.cfg 
openvpn-config-parser: [INFO] Reading file "client1.cfg"
openvpn-config-parser: [WARNING] IGNORING "status" "openvpn-status.log"
openvpn-config-parser: [WARNING] IGNORING "dh" "dh2048.pem"
openvpn-config-parser: [INFO] Reading file "ca.public.certificate"
openvpn-config-parser: [DEBUG] x509 cert: CA
openvpn-config-parser: [INFO] Reading file "client1.public.certificate"
openvpn-config-parser: [DEBUG] x509 cert: cert
openvpn-config-parser: [INFO] Reading file "client1.secret.key"
openvpn-config-parser: [ERROR] not a valid client config: does not have user/password

tls-client
dev tun
persist-key
persist-tun
topology subnet
#port 1194
remote localhost 1194
proto udp
keepalive 10 120
ca ca.public.certificate
cert client1.public.certificate
key client1.secret.key
#user ovpn
#group ovpn
#auth SHA512
tls-version-min 1.2
#tls-cipher TLS-DHE-RSA-WITH-AES-256-GCM-SHA384:TLS-DHE-RSA-WITH-AES-256-CBC-SHA256
#ncp-ciphers AES-256-GCM:AES-256-CBC
status openvpn-status.log
#log /var/log/openvpn-assym.log
verb 3
dh dh2048.pem

cfcs:
- This commit also ensures that the client specified a 'cipher' since if they did not, we end up asserting false in the 'mtu' function in src/state.ml
Fixes #24

### DIFF
--- a/src/config.ml
+++ b/src/config.ml
@@ -146,8 +146,19 @@ module Conf_map = struct
          | None | Some `Server -> Error "is not a TLS client"
          | Some `Client -> Ok ()) >>= fun () ->
         let _todo = ensure_not in
-        ensure_mem Auth_user_pass "does not have user/password"
-        (* ^-- TODO or has client certificate ? see options.c:--cert/--key,--pkcs12,or--auth-user-pass *)
+        begin match find Auth_user_pass t,
+                    find Tls_cert t,
+                    find Tls_key t
+          with
+          | None, None, None -> Error "does not have user/password"
+          | (None | Some _), Some _, None -> Error "cert but no key given"
+          | (None |Some _), None, Some _ -> Error "key but no cert given"
+          | None,   Some _, Some _
+          | Some _, None  , None
+          | Some _, Some _, Some _ -> Ok ()
+          (* ^-- TODO or has -pkcs12 *)
+        end >>= fun () ->
+        ensure_mem Cipher "client must specify 'cipher AES-256-CBC'"
         >>= fun () ->
         (if mem Cipher t && get Cipher t <> "AES-256-CBC" then
            Error "currently only supported Cipher is 'AES-256-CBC'"
--- a/src/config.ml
+++ b/src/config.ml
@@ -146,16 +146,14 @@ module Conf_map = struct
          | None | Some `Server -> Error "is not a TLS client"
          | Some `Client -> Ok ()) >>= fun () ->
         let _todo = ensure_not in
-        begin match find Auth_user_pass t,
-                    find Tls_cert t,
-                    find Tls_key t
-          with
-          | None, None, None -> Error "does not have user/password"
-          | (None | Some _), Some _, None -> Error "cert but no key given"
-          | (None |Some _), None, Some _ -> Error "key but no cert given"
-          | None,   Some _, Some _
-          | Some _, None  , None
-          | Some _, Some _, Some _ -> Ok ()
+        begin match find Auth_user_pass t, find Tls_cert t, find Tls_key t with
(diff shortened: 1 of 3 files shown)


src/config.ml
```diff
@@ -146,8 +146,19 @@ module Conf_map = struct
          | None | Some `Server -> Error "is not a TLS client"
          | Some `Client -> Ok ()) >>= fun () ->
         let _todo = ensure_not in
-        ensure_mem Auth_user_pass "does not have user/password"
-        (* ^-- TODO or has client certificate ? see options.c:--cert/--key,--pkcs12,or--auth-user-pass *)
+        begin match find Auth_user_pass t,
+                    find Tls_cert t,
+                    find Tls_key t
+          with
```
hannesm: my intuition would be that there are two cases that are fine: `Auth_user_pass` is `Some _` OR _both_ `Tls_cert` and `Tls_key` are provided. I would not know what to do if all three are provided (does it make sense to authenticate with a TLS cert + key and user/password? - does anyone do that?)

hannesm:
see #32, I'd be in favour to combine `Tls_key` and `Tls_cert` into a single configuration value.

hannesm:
I tried with Tls_key and Tls_cert as a single value, and it makes the configuration parser more cumbersome (due to inline vs external, and a similar dev/dev-type mess)... let's defer this for now, maybe we can come up with a cleaner solution.

I pushed 626460d which addresses the concerns above, please check whether you agree with them, and merge.

src/openvpn.mli
```diff
@@ -179,9 +179,6 @@ module Config : sig
         Indirectly also by [client]. *)
 
     | Tls_key      : X509.Private_key.t k
-    (** TODO Tls_key : X509.private_key * [`Incoming|`Outgoing] k
-        --key-direction governs this for inlined files
-        see comment in {!a_key} *)
```
hannesm: FWIW, this comment is wrong, it is about `--tls-auth` where we respect the direction (incoming/outgoing) already.

src/config.ml
```diff
@@ -146,10 +146,19 @@ module Conf_map = struct
...
+          | Some _, Some _, Some _  ->
+            Error "both user/pass and TLS certificate provided"
+          | _ ->
+            Error "invalid combination of user/password and TLS key and certificate"
+          (* ^-- TODO or has -pkcs12 *)
+        end >>= fun () ->
+        ensure_mem Cipher "client must specify 'cipher AES-256-CBC'"
         >>= fun () ->
-        (if mem Cipher t && get Cipher t <> "AES-256-CBC" then
+        (if get Cipher t <> "AES-256-CBC" then
```
cfcs: This (`if get Cipher t <> "AES-256-CBC" then`) now raises an exception if there is no `Cipher`.
With this patch that is currently guaranteed by the `ensure_mem` above, but openvpn files do not require `cipher` (they will pick a default somehow, for clients I think they use the cipher picked by the server in this case), so I think we should keep the two separate.
The reason for having two checks is that `current only supported Cipher is ...` is a bit confusing when the user has no `cipher` in their original config file from OpenVPN.

hannesm: ok, reverted this change in d129dfc

PR #37, opened 2020-03-28
Static pre-shared key mode (for client only so far), add a mirage-router unikernel

### LINKED ISSUE #11: static key mode
openvpn has a "simple, static key mode" -- where no ephemeral handshake is done, but pre-shared keys are used (requires an established secure channel between both parties). this means: no TLS handshake, ... . how this is implemented is:
- data is just sent to the remote, encrypted and authenticated (i.e. 20 bytes hash, 16 bytes iv, rest data)
  - data contains packed id (4 byte), time_t (4 byte), 0xfa (if no lzo), followed by the actual data
- for configuration, `OCC` (the OpenVPN configuration protocol) is used (data with 16 byte hardcoded cookie, followed by single byte command, and (optional) payload). here, a client may request the server config, discover MTU, ... (see occ.h in openvpn) -- which appears to be optional (since they broke it at 1.5 beta 8)

TL;DR: an implementation of static keys is pretty straightforward, esp. neglecting OCC. Of course, supporting OCC may be worth doing (some - such as mtu and load test and exit - may as well be used for OpenVPN-with-TLS-handshake). The question (for @cfcs) is mainly whether we want support for this, and if so, whether printng a big warning is sufficient (or should we introduce a "--i-realy-dont-need-forward-secrecy" command line argument)?

hannesm:
this fixes #11

it is atm only the client that is implemented, and only udp mode so far. there's as well some code duplication between `handle_client` and `handle_static_client` (in respect to resolving and connecting state machine) which should be cleaned up before merge. The `compression` is set to `true` (should be looked up in the config). It may be worth to support OCC (as mentioned in #11), though openvpn seems to deal fine without (ilkely makes a difference if configuration mismatches or for MTU adjustments).

I tested with a openvpn server running on my laptop, and a `mirage-client` unikernel running as virtual machine. I can successfully ping (ICMP echo request & reply) from my laptop the unikernel on its openvpn-tunneled IP address. The unikernel replies with ICMP echo reply.


hannesm:
this is getting better:
- more code is shared for the resolve + connect state machine logic between tls and static key client (there may be more to clean up in the future, but e.g. the timeout logic is slightly subtle in respect to channels and protocols)
- the `compress` is figured out from configuration, not hardcoded to `true`
- missing is tcp support (though `session.protocol` is already enhanced with the right one -- should be simple enough to add / consume the 2-byte prefix if tcp is used (and potentially fill linger)

I used the previous version (morally cea472f) for ~48 hours without an issue (on my laptop, connected via wlan and a DSL uplink which is flaky -- it worked throughout disconnects and new IP address very fine), and am now running 1de4cb2 since ~1 hour without issues. I plan to deploy 1de4cb2 on my server.

so, feedback (with the above described limitations in mind) would be great on this PR @cfcs :) while working on this, I discovered our `incoming_data` does not check for monotonicity / linearity of the packet ids, should it? i.e. at the moment it seems we're open for replay attacks of the data packets (does that matter?) -- if the answer is yes, we should, the next question would be: should we allow a window, or only accept strictly increasing numbers there? i.e. if data packets arrive out of order (1, 3, 2), should 3 be rejected? should 2 be rejected?

src/engine.ml
```diff
@@ -648,26 +694,27 @@ let timer state =
...
+let incoming_data ?(ts = false) err (ctx : keys) compress data =
   (* ok, from the spec: hmac(explicit iv, encrypted envelope) ++ explicit iv ++ encrypted envelope *)
   let hmac, data = Cstruct.split data Packet.hmac_len in
   let hmac' = Mirage_crypto.Hash.SHA1.hmac ~key:ctx.their_hmac data in
   guard (Cstruct.equal hmac hmac') (err hmac') >>= fun () ->
   let iv, data = Cstruct.split data 16 in
   let dec = Mirage_crypto.Cipher_block.AES.CBC.decrypt ~key:ctx.their_key ~iv data in
   (* dec is: uint32 packet id followed by (lzo-compressed) data and padding *)
-  let hdr_len = 4 + if compress then 1 else 0 in
+  let hdr_len = 4 + (if compress then 1 else 0) + (if ts then 4 else 0) in
```
cfcs, line 705:
```suggestion
  let hdr_len = 4 + (if ts then 4 else 0) + (if compress then 1 else 0) in
```
Since the timestamp comes before the compression tag, this ordering makes more sense to my brain :)

src/engine.ml
```diff
@@ -535,16 +567,18 @@ let unpad block_size cs =
...
 
-let data_out (ctx : keys) compress protocol rng key data =
-  (* output is: packed_id 0xfa data, then wrap openvpn partial header
+let out ?ts (ctx : keys) compress rng data =
+  (* output is: packed_id <ts_opt> 0xfa data, then wrap openvpn partial header
      ~~> well, actually take a random IV, pad and encrypt,
      ~~> prepend IV to encrrypted data
      --> hmac and prepend hash *)
-  let hdr = Cstruct.create (4 + if compress then 1 else 0) in
+  let hdr_len = 4 + (if compress then 1 else 0) + (match ts with None -> 0 | Some _ -> 4) in
```
cfcs, line 575:
```suggestion
  let hdr_len = 4 + (match ts with None -> 0 | Some _ -> 4) + (if compress then 1 else 0) in
```
Since the timestamp comes before the compression tag, this ordering makes more sense to my brain :)

cfcs:
Here's an (untested) attempt at a space-efficient sliding window algorithm to recognize repeated packet ids. The window size is `(Sys.int_size -1)` (so `62`). No idea if that's sufficient.

```ocaml
module Sliding_window : sig
  type t
  val make : int -> t
  val add : int -> t -> (t, string) result
  val mem : t -> int -> bool
  val pp : Format.formatter -> t -> unit
end = struct
  type t = { window: int; counter: int; }
  let make counter = { window = 1; counter }
  let add n {window; counter} =
    if n <= counter then begin
      let diff = counter -n in
      if diff >= Sys.int_size
      then Error "n not in sliding window"
      else Ok {window = window lor (1 lsl diff); counter}
    end else begin (* counter > n; always succeeds *)
      let diff = n - counter in
      if diff >= Sys.int_size
      then Ok (make n) (* new window *)
      else Ok {window = (window lsl diff) lor 1; counter = n}
    end

  let pp ppf {window;counter} =
    Fmt.pf ppf "{ window = %d; counter = %d }" window counter

  let mem {window;counter} n =
    let diff = counter - n in
    if counter < n || diff > Sys.int_size
    then false
    else 0 <> window land (1 lsl diff)
end
```

mirage-router/unikernel.ml
```diff
@@ -0,0 +1,148 @@
...
+    and forward_packet ip_hdr payload =
+      let c, pkt = Fragments.process t.private_fragments (M.elapsed_ns ()) ip_hdr payload in
+      t.private_fragments <- c;
+      Logs.info (fun m -> m "%B forwarding packet %a"
+                    (match pkt with None -> false | Some _ -> true)
+                    Ipv4_packet.pp ip_hdr);
+      match pkt with
+      | None -> Lwt.return_unit
+      | Some (hdr, pay) ->
+        (* we need to check mtu and potentially fragment *)
```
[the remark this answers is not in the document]
hannesm: I plan to do this today. `forward_packet` is called when a packet on the private network interface is received which should be sent over the tunnel. some lines above, we push it through the reassembly cache for the private network -- this means the full packet is now recovered (if you send a huge ping, this is completely reassembled at this point). Now, the OpenVPN tunnel likely has a smaller MTU (there's mac and padding overhead) than the private network (usually 1500 bytes) one, thus the payload needs to be cut into pieces.

An underlying issue is that some IP packets should not be fragmented, but instead an error should be sent to the sender (this is crucial for path MTU discovery, and needs to be handled here).

For the OpenVPN tunnel, there's in `Openvpn_mirage.Make()` a `  val mtu : t -> int` to be used here.

src/engine.ml
```diff
@@ -440,6 +471,7 @@ let incoming_control is_not_taken config rng state session channel now ts key op
     (est, config', session, ch'', outs)
   | Server _ ->
     incoming_control_server is_not_taken config rng session channel now ts key op data
+  | Client_static _ -> assert false
```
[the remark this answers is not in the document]
hannesm: good idea, if the API allows we may even use an `Error _`.

src/engine.ml
```diff
@@ -535,16 +567,18 @@ let unpad block_size cs =
   else
     Error (`Msg "bad padding")
 
-let data_out (ctx : keys) compress protocol rng key data =
-  (* output is: packed_id 0xfa data, then wrap openvpn partial header
+let out ?ts (ctx : keys) compress rng data =
```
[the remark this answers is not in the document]
hannesm: yeah, so in pre-shared static key mode (`secret yyy`), the wire layout is:
`<packet id> <timestamp> <data>`

while in tls (forward-secure) mode the layout is:
`<packet id> <data>`

the reason is to limit replay attacks as far as I understand. since the wire format and encryption is similar enough, I think it makes sense to use the same function. The timestamp is actually a unix time_t (i.e. seconds since epoch) as 32 bit integer (yuck, this will wrap in 2038). Now, the decode / decrypt and encode / encrypt functions do not take the entire `State.t` (which would allow a match and use the `t.now` directly), but only the components needed. I agree that the above API is fragile and should be better documented (luckily it is not exposed, but only internal).

hannesm:
about the replay protection: 
- for message-id / control packets, we do retransmissions (and preserve in State.transport a `IM.t` (which is a map keyed by Int32)) -- since control packets are ACKed and need to be retransmitted (see `expected_packet` in Engine), here we only accept the very next ID (i.e. no out-of-order, no holes) -- this is a deficiency which we should overcome (esp. important in UDP mode) 
- for packet-id / data packets, there's no ACK / retransmission. This means we don't need to store the packets sent out, and can use the sliding window implementation you suggested. I'd expect this is only affecting UDP as well (the C OpenVPN implementation is single-threaded, and I'd expect even a multi-threaded implementation to ensure sending packets in-order -- frivpn is multi-threaded, but only use a single thread for sending and ensures in-order).

The underlying question is still: how large should the window be? 3? 5? And given the above `<packet-id> <timestamp>` vs `<packet-id>`, what should the delta for the timestamp be? (this is actually related to the control packets, which include a timestamp as well, and there's a todo in the code about what to do with the timestamp)...

mirage-router/unikernel.ml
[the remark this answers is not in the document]
hannesm: after thinking a bit harder, since this is a gateway, to behave nicely in a network, the TTL field should be reduced by one, and in case the TTL turns 0, another ICMP error should be sent to the sender (instead of forwarding the packet) -- I'll implement this as well. maybe one day we should put such code in a different library and reuse it across implementations (I suspect the NAT library does something along the lines already).

src/engine.ml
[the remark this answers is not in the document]
hannesm: I revised the code in 73f4815, and while at it removed other `assert` statements from engine.ml. let me know if that commit is fine with you.

src/engine.ml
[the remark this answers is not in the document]
hannesm: done in f06621f

mirage-router/unikernel.ml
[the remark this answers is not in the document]
hannesm: done in ed5d2b6

hannesm:
remaining is implementing the replay protection -- and further testing

there's a bit of asymmetry: the `I.write` is done in a separate Lwt task (to allow that task to wait for ARP timeouts, this is crucial to retain), but the `O.write` is done in the same task (since it returns immediately (famous last words)) -- but then in `ovpn_receive` the match is not in Lwt.t, and thus I wrapped it in a Lwt.async :/

anyways, good to be squashed and merged IMHO, the replay protection is important, but can be a separate PR (I feel this is already large enough)

mirage-router/unikernel.ml
```diff
@@ -0,0 +1,271 @@
...
+      let icmp_buf = Marshal.make_cstruct icmp ~payload in
+      Cstruct.append icmp_buf payload
+    in
+    let is_icmp =
+      match Ipv4_packet.Unmarshal.int_to_protocol hdr.Ipv4_packet.proto with
+      | Some `ICMP -> true
+      | _ -> false
+    and is_first_fragment = hdr.Ipv4_packet.off land 0x1FFF = 0
+    in
+    if hdr.Ipv4_packet.ttl = 1 then (* time to live exceeded *)
```
cfcs, line 83:
```suggestion
    if hdr.Ipv4_packet.ttl <= 1 then (* time to live exceeded *)
```
Then reader doesn't have to go research if ip stack can accepts packets with `ttl = 0`

mirage-router/unikernel.ml
```diff
@@ -0,0 +1,271 @@
...
+      else begin
+        Logs.warn (fun m -> m "packet not first fragment %a, TTL exceeded"
+                        Ipv4_packet.pp hdr);
+        Error `Drop
+      end
+    else if hdr.Ipv4_packet.off land 0x4000 = 0x4000 &&
+            Cstruct.len payload > mtu then (* don't fragment set and would fragment *)
+      if is_first_fragment then
+        if is_icmp then begin
+          Logs.warn (fun m -> m "received ICMP %a with don't fragment, but would fragmentation"
```
cfcs, line 100:
```suggestion
          Logs.warn (fun m -> m "received ICMP %a with don't fragment, but would need fragmentation"
```

mirage-router/unikernel.ml
```diff
@@ -0,0 +1,271 @@
...
+      match pkt with
+      | None -> Lwt.return_unit
+      | Some (hdr, pay) ->
+        let pay_mtu = O.mtu t.ovpn - 20 in
+        match forward_or_reject hdr pay pay_mtu with
+        | Ok hdr ->
+          let hdr, fst, rest =
+            if Cstruct.len pay > pay_mtu then
+              let fst, rest = Cstruct.split pay pay_mtu in
+              let hdr = { hdr with Ipv4_packet.off = 0x2000 } in
```
cfcs, line 148: Where does `0x2000` come from here? Can we explain in a comment?

mirage-router/unikernel.ml
```diff
@@ -0,0 +1,271 @@
...
+          in
+          let hdr_cs =
+            Ipv4_packet.Marshal.make_cstruct ~payload_len:(Cstruct.len fst) hdr
+          in
+          let write_one data = O.write t.ovpn data >|= fun _ -> () in
+          write_one (Cstruct.append hdr_cs pay) >>= fun () ->
+          Lwt_list.iter_s write_one rest
+        | Error (`Icmp payload) ->
+          Lwt.async (fun () ->
+              I.write t.private_ip ~ttl:64 hdr.Ipv4_packet.src `ICMP
```
cfcs, line 161: Since we're not using path discovery, any reason this is not the max / same as here https://github.com/roburio/openvpn/pull/37/files#diff-f9e65b01a04809253309112aed0616beR234 ?

mirage-router/unikernel.ml
```diff
@@ -0,0 +1,271 @@
...
+                               proto = Ipv4_packet.Marshal.protocol_to_int `ICMP
+                    } in
+                    let payload_len = Cstruct.len pay in
+                    let hdr_cs = Ipv4_packet.Marshal.make_cstruct ~payload_len hdr in
+                    Lwt.async (fun () ->
+                      O.write t.ovpn (Cstruct.append hdr_cs pay) >|= fun _ ->
+                      ())
+                  | Error `Drop -> ()
+              else
+                Logs.warn (fun m -> m "ignoring %a (not for our network)"
```
cfcs, line 246: maybe useful to also pp `local_network` here?

src/packet.ml
```diff
@@ -48,6 +48,8 @@ let packet_id_len = 4
 
 let hmac_len = 20 (* SHA1 is what you say *)
 
+let block_size = 16
```
cfcs, line 51: Block size of what?

src/state.ml
```diff
@@ -264,24 +266,22 @@ let mtu config compress =
...
-  in
   (* padding, done on packet_id + compress + data *)
   let not_yet_padded_payload =
     4 (* packet id *) + if compress then 1 else 0
   in
   let hdrs =
-    3 (* hdr: 2 byte length, 1 byte op + key *) + bs (* IV *) + Packet.hmac_len
+    3 (* hdr: 2 byte length, 1 byte op + key *) +
+    Packet.block_size (* IV *) +
+    Packet.hmac_len
```
cfcs, line 276: For AES-256-GCM, which is now the default upstream, I believe `block size <> iv size` (12 bytes instead of 16)?

src/state.ml
```diff
@@ -264,24 +266,22 @@ let mtu config compress =
...
   in
   (* now we know: tun_mtu - hdrs is space we have for data *)
   let data = tun_mtu - hdrs in
```
cfcs, line 279: I guess this means we should ensure `tun_mtu > hdrs` in the config parser to avoid ending up with negative value here?

src/engine.ml
```diff
@@ -175,7 +204,13 @@ let incoming_tls tls data =
...
                    Fmt.(option ~none:(unit "no") Cstruct.hexdump_pp) d);
       Error (`Tls e)
-    | `Ok tls' -> Ok (tls', out, d)
+    | `Ok tls' ->
+      if expect_no_reply then
+        match out with
+        | None -> Ok (tls', out, d)
+        | Some _ -> Error (`Msg "expected no TLS reply")
+      else
+        Ok (tls', out, d)
```
cfcs, line 213: Would it maybe be cleaner to implement `no_expect_reply` as a wrapper that doesn't return `out`?
Like
```ocaml
let incoming_tls_expect_no_reply tls data = incoming_tls tls data >>= function
  | Ok (tls', None, d) -> Ok (tls', d)
  | Ok (_, Some _, _) -> Error (`Msg "expected no TLS reply")
  | Error _ as e -> e
```
Ignoring the `out` after passing the optional argument at each call site looks a bit fragile.

cfcs:
Minor suggestions, mostly about documentation and API clarity, otherwise looks good to go for me.

mirage-router/unikernel.ml
hannesm: reworded slightly different in 1af6741

mirage-router/unikernel.ml
hannesm: done in fea4b72

mirage-router/unikernel.ml
hannesm: done in d1c9579

mirage-router/unikernel.ml
hannesm: done in 742d14a

src/packet.ml
hannesm: renamed to `cipher_block_size` in 47c8b8f

src/state.ml
hannesm: maybe. out of scope -- let's fix issues when we encounter them (i.e. when someone implements multi-cipher support / AES-256-GCM).

src/state.ml
hannesm: yes, or rephrased: we should figure out a maximum / reasonable amount for `hdrs`, and check in the configuration parser that `Tun_mtu` (either in the configuration file, or transmitted via PUSH_REPLY) is way above that value. Since we're (atm?) concerned about tun interfaces only, we should in addition ensure that the remaining number is well above 20 (the IPv4 header size).

src/engine.ml
hannesm: that's a good suggestion, done in caef48b

src/state.ml
hannesm: I opened #51 to track this issue. Thanks for commenting on this function, in 56de7cf I added conditionals for static keys (where the header includes a timestamp, but excludes the 1 byte operation and key id).

PR #38, opened 2020-03-28
test for parsing server config

whohoho:


src/config.ml
```diff
@@ -1,3 +1,4 @@
+
```
cfcs: why did this need to change? :)

src/config.ml
```diff
@@ -135,6 +136,33 @@ module Conf_map = struct
...
+            Error "no tls key"
+          (* ^-- TODO or has -pkcs12 *)
+        end >>= fun () ->
+        ensure_mem Cipher "server must specify 'cipher AES-256-CBC'"
+        >>= fun () ->
+        (if mem Cipher t && get Cipher t <> "AES-256-CBC" then
+           Error "currently only supported Cipher is 'AES-256-CBC'"
+         else Ok ()) 
+      )
+
```
cfcs: Wonder if it'd make sense to have a common function to check the common subset so we avoid duplication?

src/config.ml
```diff
@@ -369,6 +397,26 @@ module Defaults = struct
...
+    |> add Handshake_window 60
+    |> add Transition_window 3600
+    |> add Tls_timeout 2
+    |> add Resolv_retry `Infinite
+    |> add Auth_retry `None
+    |> add Connect_timeout 120
+    |> add Connect_retry_max `Unlimited
+    |> add Proto (Some `Ipv4, `Tcp (Some `Server))
+    |> add Tls_mode `Server
+    |> add Server ((Ipaddr.V4.of_string_exn "10.89.0.0"), Ipaddr.V4.Prefix.of_string_exn "10.89.0.0/24")
```
cfcs: - I believe the default device should be `Dev (Tun, None)` ? Does having `Tun` as default here work when making a `tap` server?
- isn't the default port 1194? I think it should be `None, None` here to allow distinguishing between default vs literal port 1194.
- does the server only listen on ipv4 by default? no ipv6?
- Couldn't find any mention of this ip in openvpn. Seems like something we should complain about missing rather than setting by default?

src/config.ml
```diff
@@ -1455,7 +1503,7 @@ let parse_next (effect:parser_effect) initial_state : (parser_state, 'err) resul
             | [], name::_extra_devs ->
               Error (Fmt.strf
                        "[dev %S] stanza without required [dev-type]" name)
-            | _ -> Error "multiple conflicting [dev-type] stanzas present"
+            | _ -> Error "multiple (maybe conflicting) [dev-type] stanzas present"
```
cfcs: The code above aims to detect overlapping non-conflicting dev-type/dev things and explain what conflicts, so I think we should consider it a bug if there are scenarios where this happens. Do you have an example?

src/openvpn.mli
```diff
@@ -246,6 +246,22 @@ module Config : sig
   (** Parses a configuration string, looking up references to external files
       as needed. Validates the client configuration. Default client options are
       applied. *)
+
+  val parse_server : string_of_file:(string -> (string, R.msg) result) ->
+    string -> (t, [> R.msg]) result
+  (** Parses a configuration string, looking up references to external files
+      as needed. Validates the client configuration. Default client options are
+      applied. *)
```
cfcs: Validates the **client** configuration? ;)

test/config_tests.ml
```diff
@@ -47,7 +59,44 @@ testpass
...
+  |> add Transition_window 3600
+  |> add Tls_timeout 2
+  |> add Resolv_retry `Infinite
+  |> add Auth_retry `None
+  |> add Connect_timeout 120
+  |> add Connect_retry_max `Unlimited
+  |> add Proto (Some `Ipv4, `Tcp (Some `Server))
+  |> add Tls_mode `Server
+  |> add Server ((Ipaddr.V4.of_string_exn "10.89.0.0"), Ipaddr.V4.Prefix.of_string_exn "10.89.0.0/24")
+  |> add_ok_b (a_ca_payload (string_of_file "ca.public.certificate" ) : (b,'a) result)
```
cfcs: Is `ca` required in a minimal config? We don't check for it in the `Config` module.

src/config.ml
```diff
@@ -136,12 +136,56 @@ module Conf_map = struct
...
+
+  let is_valid_server_config t =
+    let ensure_mem k err = if mem k t then Ok () else Error err in
+    let ensure_not k err = if not (mem k t) then Ok () else Error err in
+    let open Rresult in
+    R.reword_error (fun err -> `Msg ("not a valid server config: " ^  err))
+      ( 
+        match is_valid_config t with
+          | Error _ -> Error "Not a valid config"
+          | Ok _ -> Ok () >>= fun () ->
```
cfcs: This needs to go before the `R.reword_error`, unless we get
`not a valid server config: not a valid config: currently only supported Cipher is 'AES-256-CBC'` which is a tad verbose :)

(same in the client config checker)

src/config.ml
```diff
@@ -136,12 +136,56 @@ module Conf_map = struct
...
+        ensure_mem Bind "does not have a bind" >>= fun()->
+        (match find Tls_mode t with
+         | None | Some `Client -> Error "is not a TLS server"
```
cfcs: maybe we should mention how to make it one here, ie the directive required.

src/config.ml
```diff
@@ -136,12 +136,56 @@ module Conf_map = struct
...
+         | Some `Server -> Ok ()) >>= fun () ->
+        let _todo = ensure_not in
+        begin match find Tls_cert t, find Tls_key t with
+          |  Some _, Some _ -> Ok ()
+          |  None, None ->
+            Error "does not have  TLS certificate and key"
+          | Some _, None -> 
+            Error "no tls certificate"
+          | None, Some _ ->
+            Error "no tls key"
```
cfcs:
```suggestion
            Error "missing tls-key"
```

src/config.ml
```diff
@@ -136,12 +136,56 @@ module Conf_map = struct
...
+        (match find Tls_mode t with
+         | None | Some `Client -> Error "is not a TLS server"
+         | Some `Server -> Ok ()) >>= fun () ->
+        let _todo = ensure_not in
+        begin match find Tls_cert t, find Tls_key t with
+          |  Some _, Some _ -> Ok ()
+          |  None, None ->
+            Error "does not have  TLS certificate and key"
+          | Some _, None -> 
+            Error "no tls certificate"
```
cfcs:
```suggestion
            Error "missing tls-cert"
```

src/config.ml
```diff
@@ -136,12 +136,56 @@ module Conf_map = struct
...
+          | Ok _ -> Ok () >>= fun () ->
+        ensure_mem Bind "does not have a bind" >>= fun()->
+        (match find Tls_mode t with
+         | None | Some `Client -> Error "is not a TLS server"
+         | Some `Server -> Ok ()) >>= fun () ->
+        let _todo = ensure_not in
+        begin match find Tls_cert t, find Tls_key t with
+          |  Some _, Some _ -> Ok ()
+          |  None, None ->
+            Error "does not have  TLS certificate and key"
```
cfcs:
```suggestion
            Error "missing tls-cert and tls-key"
```

test/config_tests.ml
```diff
@@ -1,16 +1,36 @@
...
+      Rresult.R.error_msgf
+        "this test suite does not read external files, \
+         but a config asked for: %S" path) conf
+
+let ok_cert_userpass_client () =
+  let conf = string_of_file "client-tcp-certauth-passauth.cfg" in
+  let parsed = parse_noextern_client conf in
+  ignore parsed;
+  (*FIXME*)
+  Alcotest.(check bool) "cert_userpass works" true true
```
cfcs: I'm adding a dummy comment here we can resolve when this is implemented.

src/config.ml
whohoho: changed it back to conflicting for now

hannesm:
sorry, I just got around to look into this issue. What do you think about #47 and rebasing this PR on top (I am volunteering to do this if you think it is a good idea -- it looks like the goals are partially shared between those two (in terms of testing)).

This PR adds `User : string k` into the configuration map. My general approach is to only put things into the configuration map that are actually used by the library (and put others into the `a_not_implemented` list).

Why does the Defaults.server_config contain those elements in there? It looks like a copy of the Defaults.client_config -- i.e. `Resolv_retry`, `Auth_retry` -- are they relevant for a server? Maybe we should have a Defaults.config map -- shared between client and server -- and Defaults.client_config and Defaults.server_config that extend the Defaults.config?

hannesm:
It looks like you forgot to `git add app/openvpn_server_config_parser.ml`.

src/openvpn.mli
```diff
@@ -246,6 +246,22 @@ module Config : sig
...
+
+  val a_ca_payload : string -> (b, string) result
+  (**  for test *)
+  
+  val a_cert_payload : string -> (b, string) result
+  (**  for test *)
+
+  val a_key_payload : string -> (b, string) result
+  (**  for test *)
```
hannesm: I prefer to not export these bindings. I know this means that we can "only" test a "full" configuration, which is fine with me. The amount of stuff in `Openvpn.mli` should be minimal.

hannesm: I meant any of the above, I'm more in favour to potentially replicate some code in the test library rather than have these bindings here. to me, it looks like X509.Certificate.decode_pem etc. already do what needs to be done here (apart from the static key/tls-auth hex-decoding and split into four sections).

whohoho: Ok, so i copied those functions into the test

hannesm: thanks, that looks much cleaner to me.

test/config_tests.ml
whohoho: removed it for now

hannesm:
this PR touches multiple things at once:
- test suite improvements
- adding app/open_server_config_parser
- extending the configuration parser (parse more things, validate_server_config, ..)

I'd appreciate to separate these distinct parts (maybe b and c are not too disjoint and could be a single PR with two commits). for (b), I'm wondering whether a second binary is really necessary (which shares >90% of code with the existing one), instead of extending the existing openvpn_config_parser with some code to discover whether it is a client or server configuration (by looking at Tls_mode / Proto / Remote_cert_tls / whether Remote is present -- or take a deeper look how the C openvpn discovers this).

whohoho:
https://github.com/roburio/openvpn/pull/48

hannesm:
This has been split into multiple PRs, some are already merged (since a long time), others are still open. I'll close this PR since it is now out of date and the core contributions are tracked separately.

Thanks a lot, sorry for the long delay.

PR #41, opened 2020-03-28
Support both user+password AND client cert auth

### LINKED ISSUE #39: its fine to have user/pass AND tls cert auth in client config
with this commit: 
https://github.com/roburio/openvpn/commit/d129dfcb36b54b02193da6608115d801b1b955a1

a config would be invalid if you use both user/pass and a client certificate, but that would a valid config for openvpn

cfcs:
Closes #39

### DIFF
--- a/src/config.ml
+++ b/src/config.ml
@@ -147,13 +147,12 @@ module Conf_map = struct
          | Some `Client -> Ok ()) >>= fun () ->
         let _todo = ensure_not in
         begin match find Auth_user_pass t, find Tls_cert t, find Tls_key t with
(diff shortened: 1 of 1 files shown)


hannesm:
thanks, esp. for the more in-depth discussion why it can be useful to authenticate via TLS client certificate _and_ user+pass: two-factor authentication with an OTP (see e.g. https://github.com/evgeny-gridasov/openvpn-otp) -- I suspect getting our OpenVPN implementation into shape with OTP will require some more effort (and needs someone with a setup that requires it ;)

Issue #42, opened 2020-03-28
verify-x509-name is not parsed


OtherSystems:
**Client config**
    
    client
    tls-client
    dev tun
    persist-tun
    remote vpn.example 1194 udp
    verify-x509-name vpn.example name
    resolv-retry infinite
    keepalive 10 60
    route-nopull
    route 192.168.2.0 255.255.255.0
    <ca>
    -----BEGIN CERTIFICATE-----
    ca cert
    -----END CERTIFICATE-----
    </ca>
    <cert>
    -----BEGIN CERTIFICATE-----
    cert
    -----END CERTIFICATE-----
    </cert>
    <key>
    -----BEGIN RSA PRIVATE KEY-----
    key
    -----END RSA PRIVATE KEY-----
    </key>
    persist-key
    comp-lzo yes
    verb 3
    cipher AES-256-CBC
    auth SHA256
    lport 0
    verb 4
    auth-nocache
    log /var/log/openvpn.log

**Server config**

    verb 4
    dev uvpn
    dev-type tun
    dev-node /dev/net/tun
    writepid /var/run/openvpn_tunvpn.pid
    script-security 3
    keepalive 10 60
    ping-timer-rem
    persist-tun
    persist-key
    proto udp
    cipher AES-256-CBC
    auth SHA256
    tls-server
    server 192.168.6.128 255.255.255.128
    client-config-dir /etc/openvpn/server
    local vpn.example
    lport 1194
    route-up "/usr/local/sbin/openvpn-set.pl"

hannesm:
thanks for your report, I edited the title to reflect what is the issue here. The provided configuration files are now part of our testsuite (at the moment commented out since parsing fails).

PR #47, opened 2020-03-29
test: add client configuration parse tests from sample configuration files

hannesm:
TL;DR: let's gather in `tests/sample-configuration-files` configuration files we would like to be successfully parsed. At the moment, these are based on the examples distributed with OpenVPN.

This PR adds a (parametrised) test `parse_client_configuration` which takes a filename and attempts to parse this file in `sample-configuration-files`. Only "is the parse successful" is tested, not the contents of the configuration. Due to the rise of interest/traffic in this repository, I think it is a good idea to gather (normalised = using the same secret / ta / certificate / key / password) configurations we currently support / would like to support in the future.

NB: tls-home.conf has been modified to include a `cipher` statement. By reading the man page, OpenVPN seems to switch their default cipher every other version (blowfish, AES-CBC, AES-GCM), thus I suggest to (as is atm) not support configuration files without an explicit `cipher` line (as done by #31): our goal is not to support all the sample configuration files OpenVPN contains, but rather more real-world examples used in the wild (#42, riseup, ~~ipredator~~, etc.).


test/config_tests.ml
```diff
@@ -234,6 +234,29 @@ tls-auth [inline]
...
+    let content = really_input_string fh (in_channel_length fh) in
+    close_in_noerr fh ;
+    content
+  with _ -> Alcotest.failf "Error reading file %S" filename
+
+let config_dir = "sample-configuration-files/"
+
+let parse_client_configuration name () =
+  Unix.chdir config_dir;
+  Fun.protect ~finally:(fun () -> Unix.chdir "..")
```
cfcs: Hmm, `chdir` inside a test suite seems like a risky thing to be doing since we don't know what else is writing files, can't we make do with absolute paths?

test/sample-configuration-files/client.crt
```diff
@@ -0,0 +1,103 @@
+Certificate:
+    Data:
+        Version: 3 (0x2)
```
cfcs: Hmm, surprised the parser ignores this when not prefixed by `#` ??

openvpn.opam
```diff
@@ -16,8 +16,8 @@ build: [
 ]
 
 depends: [
-  "dune" {build}
-  "ocaml" {>= "4.07.0"}
+  "dune"
+  "ocaml" {>= "4.08.0"}
```
cfcs: Why do we need to leave 4.07 behind for this changeset?

test/config_tests.ml
hannesm: since the configuration files refer to "files on the local system", it is hard to do with absolute paths.

in my perspective, the `Fun.protect` does exactly the right thing here -- ensuring that `Unix.chdir ".."` is always executed before the test is finished.

test/sample-configuration-files/client.crt
hannesm: X509.Certificate/Private_key.decode_pem only looks for `-----BEGINYYYY------ ... -----ENDYYYY-----` lines :)

which in my opinion is pretty fine (from looking into my global trust anchors (/etc/ssl/cert.pem), this is definitively the way to go)

openvpn.opam
hannesm: `Fun.protect`.

test/config_tests.ml
cfcs: Oh, this raises the question how openvpn does this - whether relative paths are relative to current execution dir or the config file itself. I would expect it to be relative to the config file (in which case we shouldn't need `chdir`).

hannesm: well, we'd need the logic for the parser to do relative to the configuration file -- which atm is not present (I tried first without `chdir`, it failed, so I came up with this) ;)

but if that's something you think should be implemented by the config parser, please feel free (you could push this to this PR if you like). this would as well get rid of `Fun.protect` and the 4.08.0 lower bound.

cfcs: (but we may need to adjust the config file parser if that's the case)

hannesm: well, the original `openvpn` actually takes files referenced in a configuration as relative to the current working directory -- i.e. `openvpn config/client.conf` (with embedded `ca ca.crt`) fails:
```bash
Options error: --ca fails with 'ca.crt': No such file or directory (errno=2)
Sun Mar 29 22:33:58 2020 us=707536 WARNING: cannot stat file 'ta.key': No such file or directory (errno=2)
```

hannesm: but openvpn has the `--cd` argument to change its working directory "before reading any files such as configuration files"... not immediately useful here since we're using it as a library..

cfcs: This means that the semantics of the config file depends on the working directory, which seems really ugly to me (ie execution the same openvpn commandline with the same arguments can lead to different results based on what the cwd is).
It also implies that e.g. our tool to read/check configuration files should also work different depending on your cwd (but with the same file).
I'd be super inclined to change that to `~/.ssh/config`-style semantics where paths are relative to the location of the file that makes the reference, so a config file can only be parsed one way.

hannesm: I did so in 4a0c60f.

hannesm:
I revised the configuration parser (lwt and tests) call sites to treat relative paths as relative to the configuration file, and documented this in the README. This does _not_ affect the MirageOS layer since there we don't consider any external files. Would be good to get this merged if you're fine with the changes @cfcs.

hannesm:
I included #48 here (slightly amended), and extended the test to check the contents of parsed client configurations.

hannesm:
this (a) checks the contents of configurations as well (b) includes @whohoho #48 (but commented out the server side configuration testing, which requires #50), and (c) fixes the (mostly client) configurations so they could be parsed (external referenced files exist now).

test/config_tests.ml
```diff
@@ -1,16 +1,79 @@
+
+let a_x509_cert_payload ctx constructor str =
+  Logs.debug (fun m -> m "x509 cert: %s" ctx);
+  match X509.Certificate.decode_pem (Cstruct.of_string str) with
+  | Ok cert -> constructor cert
+  | Error (`Msg msg) -> Alcotest.failf "%s: invalid certificate: %s" ctx msg
+
+let a_ca_payload str =
+  let open Openvpn.Config in
+  a_x509_cert_payload "CA" (fun c -> B(Ca,c)) str
```
cfcs:
```suggestion
  a_x509_cert_payload "ca" (fun c -> B(Ca,c)) str
```

test/config_tests.ml
```diff
@@ -47,7 +109,40 @@ testpass
...
+
+let minimal_server_config =
+  let open Openvpn.Config in
+  empty
+  |> add Dev (`Tun ,(Some "tunnel"))
+  |> add Ping_interval `Not_configured
+  |> add Cipher "AES-256-CBC"
+  |> add Ping_timeout (`Restart 120)
+  |> add Renegotiate_seconds 3600
+  |> add Bind (Some (Some 1195, None))
```
cfcs: See my review on #38: https://github.com/roburio/openvpn/pull/38#discussion_r399712089
#38 has fixes for most of this, e.g. https://github.com/roburio/openvpn/pull/38/commits/64ebf6f6c73ac828fc91191fe996ded5f5b313e0
and

hannesm: your comment in #38 is about `Config.Defaults.server` (this is Config_test.minimal_server_config -- and actually unused since we don't have a parse_server_config yet) -- I'd be happy to get rid of the minimal_*_config in test and reuse Config.Defaults.yy in a subsequent PR. in the meantime, I just deleted the commented out and unused code in 17a995c5a15d83c6952701138e1ea3104a4d8539.

PR #49, opened 2020-03-29
openvpn_server_config_parser.exe

whohoho:


cfcs:
Wouldn't it be nicer to just have one config parser, and have that emit a comment at the top of the output that describes whether it's a client or server config?

hannesm:
@cfcs I agree, from my comment in #38:

I'm wondering whether a second binary is really necessary (which shares >90% of code with the existing one), instead of extending the existing `openvpn_config_parser` with some code to discover whether it is a client or server configuration (by looking at Tls_mode / Proto / Remote_cert_tls / whether Remote is present -- or take a deeper look how the C openvpn discovers this).

whohoho:
> Wouldn't it be nicer to just have one config parser, and have that emit a comment at the top of the output that describes whether it's a client or server config?

i agree

whohoho:
for now i have:

+  type config_type =
+    | Tls_tcp_server
+    | Tls_tcp_client
+    | Tls_udp_server
+    | Tls_udp_client
+    | P2p_udp
+    | P2p_tcp_server
+    | P2p_tcp_client
+    | Invalid

Does that make sense?

hannesm:
@whohoho I'm not sure what your plan with `type config_type` is. I'd keep it simple for now, and only distinguish "is this a `Server` or a `Client` configuration". Depending on that, call `valid_client_config` or `valid_server_config` (in there, it may make sense to distinguish the configuration further, i.e. no tls, ...).

whohoho:
> @whohoho I'm not sure what your plan with `type config_type` is. I'd keep it simple for now, and only distinguish "is this a `Server` or a `Client` configuration". Depending on that, call `valid_client_config` or `valid_server_config` (in there, it may make sense to distinguish the configuration further, i.e. no tls, ...).

As far as i understand, openvpn initially had a static key mode in p2p (so no client no server)
Then later they added the p2mp mode , which uses tls.

So in the p2p mode there is no concept of a server (except when using tcp, when the one opening the listening socket is the server)

hannesm:
hmm, ok. but does this matter? This repository / package is not meant to re-iterate all the OpenVPN decisions they made. From my perspective, there's always a client (the one with a `Remote` in its configuration which starts establishing a tunnel), and a server (the one listening for incoming requests and managing IP addresses). I don't see much point in supporting the old `p2p` mode, which I haven't seen being used at all.

whohoho:
> hmm, ok. but does this matter? This repository / package is not meant to re-iterate all the OpenVPN decisions they made. From my perspective, there's always a client (the one with a `Remote` in its configuration which starts establishing a tunnel), and a server (the one listening for incoming requests and managing IP addresses). I don't see much point in supporting the old `p2p` mode, which I haven't seen being used at all.

Isn't https://github.com/roburio/openvpn/pull/37 adding support for static key mode?

hannesm:
> Isn't #37 adding support for static key mode?

yes, this adds static-key mode (client side), based on a configuration I actually have and use.

whohoho:
> > Isn't #37 adding support for static key mode?
> 
> yes, this adds static-key mode (client side), based on a configuration I actually have and use.

there is no client or server side in static key mode (when using udp)

This is where openvpn parses the config options:
https://github.com/OpenVPN/openvpn/blob/0d94d433438f239ff7cf0749f765a503c698f5e8/src/openvpn/options.c#L1977

hannesm:
I feel I'm repeating myself: from a programmatic point of view, I distinguish between a client (which has a `Remote` configured and is responsible to send the first packet) and a server (which sits there, listens for incoming connections). This is crucial to figure out which state machine to use (resolve / connect OR bind / listen).

Even in static key mode (as far as my example configurations are, also what is shipped with the C openvpn (static-home.conf / static-office.conf -- only one has a `remote` line)), there's this concept of a client and a server.

whohoho:
ah, mm, so i can say that p2p mode without remote is a server, and with a remote is a client. Which means that you can then have 2 clients talking to each other.

whohoho:
cat p2p-1.cfg 
remote localhost 1235
local localhost
lport 1234 
dev tun
ifconfig 10.8.0.1 10.8.0.2
secret ta.key

cat p2p-1.cfg 
remote localhost 1235
local localhost
lport 1234 
dev tun
ifconfig 10.8.0.1 10.8.0.2
secret ta.key

cat p2p-1-noremote.cfg 
#remote localhost 1235
local localhost
lport 1234 
dev tun
ifconfig 10.8.0.1 10.8.0.2
secret ta.key

whohoho:
cat p2p-1-nobind.cfg 
remote localhost 1235
dev tun
ifconfig 10.8.0.1 10.8.0.2
secret ta.key

openvpn p2p-1-nobind.cfg 
Mon Mar 30 15:23:53 2020 setsockopt(IPV6_V6ONLY=0)
Mon Mar 30 15:23:53 2020 UDP link local (bound): [AF_INET6][undef]:1194
Mon Mar 30 15:23:53 2020 UDP link remote: [AF_INET6]::1:1235

hannesm:
I clicked to target the master branch, now some unrelated commits are appearing here (sorry :/).
Would you mind to rebase, and address the comments above and what we discussed on IRC?

To rewrap, I think:
- integrate this into the `openvpn_config_parser`
- use the presence of `Remote` to investigate whether this is a client or a server configuration
- this may need to come after #50 (or be included) to be able to validate the server configuration

Thanks :)

hannesm:
Thanks for your work on this. It is slightly out of date with the current state of this repository. We'll revise what to do for the server side soon.

PR #50, opened 2020-03-29
server config checks in config.ml

whohoho:


hannesm:
IMHO this is fine to be merged, only some minor nit:
Why does the Defaults.server_config contain those elements in there? It looks like a copy of the Defaults.client_config -- i.e. Resolv_retry, Auth_retry -- are they relevant for a server? Maybe we should have a Defaults.config map -- shared between client and server -- and Defaults.client_config and Defaults.server_config that extend the Defaults.config? Should `Tls_mode Server` really be the default (there's no `Tls_mode Client` in the Defaults.client_config)?

whohoho:
> IMHO this is fine to be merged, only some minor nit:
> Why does the Defaults.server_config contain those elements in there? It looks like a copy of the Defaults.client_config -- i.e. Resolv_retry, Auth_retry -- are they relevant for a server? Maybe we should have a Defaults.config map -- shared between client and server -- and Defaults.client_config and Defaults.server_config that extend the Defaults.config? Should `Tls_mode Server` really be the default (there's no `Tls_mode Client` in the Defaults.client_config)?

ah, yes, default mode is p2p
       --mode m
              Set OpenVPN major mode.  By default, OpenVPN runs in point-to-point mode ("p2p").  OpenVPN 2.0 introduces a new mode
              ("server") which implements a multi-client server capability.

I made defaults for client and server the same now.

hannesm:
> I made defaults for client and server the same now.

repeating myself: i.e. `Resolv_retry`, `Auth_retry` -- are they relevant for a server? I doubt the approach "client and server are equal" actually works in practise, and would suggest to have a "Defaults.config" _and_ "server / client", where the former is extended by client and server.

whohoho:
> > I made defaults for client and server the same now.
> 
> repeating myself: i.e. `Resolv_retry`, `Auth_retry` -- are they relevant for a server? I doubt the approach "client and server are equal" actually works in practise, and would suggest to have a "Defaults.config" _and_ "server / client", where the former is extended by client and server.

No, they are not used by server (but are allowed to be in config file), but indeed it makes sense to not have those defaults for when you want to write a config file.

hannesm:
I clicked here to target the master branch, and now some commits from the other branch are appearing :/ -- would you mind to rebase your branch on the current master; and address the remarks above? thanks!

hannesm:
from re-reading the current code, I figured there's a TODO with some items I considered back then being important to have in a server config -- https://github.com/roburio/openvpn/blob/c2c37f3b0e15a4d19263b85d82a556fc2555301a/src/engine.ml#L82-L84

hannesm:
Thanks a lot, I took the liberty to rebase and adapt in #101 the core of this PR.

PR #54, opened 2020-05-17
openvpn_config_parser.exe: pad output to block size

cfcs:
to enable use with unikernels through block dev api

output similar to:

```
#############          
# IMPORTANT:
#  This OpenVPN configuration file has been padded with the comment below to
#  ensure alignment on 512-byte boundaries for block device compatibility.
#  That is a requirement for the MirageOS unikernels.
#  If you modify it, please verify that the output of
#       wc -c THIS.FILE
#  is divisible by 512.
#############

###############################################
############################################################################
############################################################################
############################################################################
############################################################################

client
auth-retry nointeract
auth-user-pass [inline]
.......
```

### DIFF
--- a/app/openvpn_config_parser.ml
+++ b/app/openvpn_config_parser.ml
@@ -22,7 +22,42 @@ let read_config_file fn =
   | Ok str -> parse_client ~string_of_file str
   | Error _ as e -> e
 
+let alignment_header =
+  {|#############
+# IMPORTANT:
+#  This OpenVPN configuration file has been padded with the comment below to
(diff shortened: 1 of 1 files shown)


cfcs:
CI fails with:
```
- File "src/config.ml", line 634, characters 4-19:
- Error: This pattern matches values of type ('a, 'b) result
-        but a pattern was expected which matches values of type
-          consume:Angstrom.Consume.t ->
-          (Cstruct.t * Cstruct.t * Cstruct.t * Cstruct.t, string) result
```
Looks like we need to adapt to new Angstrom upstream.

PR #61, opened 2023-03-17
Rename to MirageVPN

reynir:
The repository would need to be renamed as well.

Renaming the modules was maybe not necessary.

I have as well renamed the log source module in `Miragevpn.Config` from `Logs` to `Log`. I had accidentally renamed openvpn.mli to miragevpn***g***.mli which then exposed the Config.Logs module. Otherwise this PR is just renaming.

I have not touched the diagrams and they may need updating as well.


dinosaure (approved):

hannesm:
thanks a lot

PR #64, opened 2023-09-08
Apply ocamlformat.0.26.0

dinosaure:
A gently upgrade to use `ocamlformat.0.26.0`.


dinosaure:
Note that currently the code does not compile but I have a patch which fix some `Lwt.catch` in the client application.

hannesm:
would you mind to put that rev into a new `.git-blame-ignore-revs` file at the root (see https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view)

Also, I wonder why OCaml CI doesn't run here... maybe we should setup a simple GitHub action or cirrus CI?

dinosaure:
About `ocaml-ci`, it's related to https://github.com/ocurrent/ocaml-ci/pull/867

hannesm:
build has been fixed in #66, please rebase and add the git-blame-ignore-revs file :) thanks!

dinosaure:
This PR was rebased :+1:.

hannesm:
thanks, I added a .git-blame-ignore-revs file myself, pushed, and merged.

PR #67, opened 2023-09-12
Use failwith instead of Lwt.fail_with

reynir:
`Lwt.fail` and `Lwt.fail_with` do not produce nice stack traces, and in most cases directly raising the exception (or using `failwith`) is preferable. The exceptions here are not very interesting and should perhaps be replaced by printf/log message followed by `exit N`.


hannesm:
thanks

PR #68, opened 2023-09-12
Shuffle remotes if remote-random is present

reynir:
This is a first attempt at implementing `remote-random`. A helper function `get_remotes` is added that checks for `remote-random` and shuffles the remotes if required.

This is a draft PR as I need to verify that the remotes are not shuffled more than once. As this code is brittle in this way I may reconsider the PR.

### DIFF
--- a/src/engine.ml
+++ b/src/engine.ml
@@ -85,7 +85,8 @@ let get_remotes rng config =
 
 let client config ts now rng =
   let current_ts = ts () in
-  (match get_remotes rng config with
+  let remotes = get_remotes rng config in
+  (match remotes with
   | (`Domain (name, ip_version), _port, _proto) :: _ ->
       Ok (`Resolve (name, ip_version), Resolving (0, current_ts, 0))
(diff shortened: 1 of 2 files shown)


dinosaure (approved):

hannesm:
looks fine to me :) thanks.

reynir:
From studying the code a bit closer I think this will not work:
- `Miragevpn.handle` is called with a new event, such as `` `Resolved``.
- `Miragevpn.handle` in turn calls `Miragevpn.resolve_connect_client` with the event. `resolve_connect_client` calls `get_remotes` which returns a shuffled remotes list (if `remote-random` is specified).
- When the event is `` `Resolved`` the current remote is looked up by the index stored in the client state to get the port and protocol (tcp/udp) - but the remotes list has been shuffled since that index was passed in the client state and may refer to a different remote now.

The client state may need to be extended with a list of remotes.

reynir:
Closing this in favor of #69. Please reopen if you disagree.

PR #69, opened 2023-09-18
Shuffle remotes

reynir:
This is another take on #68. If `remote-random` is present the config object is (immutably) updated with a shuffled remotes list when creating a connection in `Miragevpn.client`.

This change is simpler and touches less code.


dinosaure:
LGTM, if you are confident about this change, let's merge.

hannesm:
thanks

PR #71, opened 2023-09-19
Replace our own implementation of lzo by decompress.lzo

dinosaure:
decompress.lzo implements also a compressor which can help about the implementation of a VPN server. Tests are kept to check that everything works even with the replacement. The directory minilzo is removed (because unused) and lzo_fuzzable.ml was updated even if decompress.lzo has its own fuzzer.


dinosaure:
Currently, we use the _upstream_ version of `decompress` but a release can be done (a minor one, the interface does not change at all!).

hannesm:
Thanks. Indeed I prefer to have a decompress release and specify here in opam only the lower bound :)

reynir (approved):

reynir:
Thanks! The change seems straight forward and with decompress released I think this is safe to merge.

PR #72, opened 2023-09-19
Remove `rresult` package from the core library

dinosaure:
Unikernels are not updated yet. I preferred to locally open `Result.Infix` in some parts of the code (specially `engine.ml`). The core library provide a little `Result` module which implements the `Infix` submodule. Due to `wrapped true` (into the `dune` file), I don't think that such module will be an issue for an external point of view.


hannesm:
looks good to me. I also appreciate the local opens and the Infix module.

reynir (approved):

PR #73, opened 2023-09-25
remove rresult, also from unikernels

hannesm:
in addition to #72 this removes rresult across the repo


hannesm:
this should be fine now... I noticed some changes needed for mirage-nat unikernel (due to mirage-nat API changes some time ago). I'm glad the unikernels are now being built within CI

dinosaure:
Thanks!

PR #74, opened 2023-09-25
remove astring dependency

hannesm:


src/engine.ml
```diff
@@ -398,11 +398,16 @@ let maybe_push_reply config = function
...
         Logs.info (fun m -> m "push request sent, received TLS payload %S" str);
-        match Astring.String.cut ~sep:"PUSH_REPLY" str with
-        | Some ("", opts) -> Config.merge_push_reply config opts
-        | _ ->
-            Error
-              (`Msg (Fmt.str "push request expected push_reply, got %S" str)))
+        let p_r = "PUSH_REPLY" in
+        let p_r_len = String.length p_r in
+        if String.length str >= p_r_len &&
+           String.(equal (sub str 0 p_r_len) p_r)
```
reynir, line 404: Since 4.13.0 we can use `String.starts_with`:
```suggestion
        if String.starts_with str ~prexix:p_r
```

test/config_tests.ml
```diff
@@ -29,16 +29,20 @@ let a_inline_payload str =
   let rec content acc collect = function
     | [] -> List.rev acc
     | hd :: tl when collect ->
-        if Astring.String.is_prefix ~affix:"-----END" hd then
+        let aff = "-----END" in
+        let aff_len = String.length aff in
+        if String.length hd >= aff_len && String.(equal (sub hd 0 aff_len) aff) then
```
reynir, line 34:
```suggestion
        if String.starts_with ~prefix:"-----END" hd then
```

test/config_tests.ml
```diff
@@ -29,16 +29,20 @@ let a_inline_payload str =
...
           content acc false tl
         else content (hd :: acc) true tl
     | hd :: tl ->
-        if Astring.String.is_prefix ~affix:"-----BEGIN" hd then
+        let aff = "-----BEGIN" in
+        let aff_len = String.length aff in
+        if String.length hd >= aff_len && String.(equal (sub hd 0 aff_len) aff) then
```
reynir, line 40:
```suggestion
        if String.starts_with ~prefix:"-----BEGIN" hd then
```

reynir (approved):
Looks good to me. If we require OCaml>=4.13.0 we can use `String.starts_with`.

src/config.ml
```diff
@@ -1813,9 +1814,10 @@ let client_merge_server_config client server_str =
       m "Config.client_merge_server_config @[<v>is not implemented@ %S@]"
         server_str);
   Ok client
-(*let str = Astring.String.(concat ~sep:"\n" (cuts ~sep:"," server_str)) in
-  parse ~string_of_file:(fun fn ->
-      Result.error_msgf "Server requested client to read %S" fn)
-    str >>= valid_server_options ~client >>| fun () -> client*)
+(* String.split_on_char ',' server_str
```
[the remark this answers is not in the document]
dinosaure: We will do that in another PR to be conservative. We don't know yet precisely why this code is commented.

dinosaure:
Thanks!

PR #83, opened 2023-10-04
Add support for --tls-cipher and --tls-ciphersuite

### LINKED ISSUE #45: tls-cipher is not parsed
openvpn_server_config_parser.exe windows-riseup-client.cfg 
openvpn_server_config_parser.exe: [INFO] Reading file "windows-riseup-client.cfg"
openvpn_server_config_parser.exe: [ERROR] : Error at byte offset 988: "tls-cipher TLS-DHE-RSA-WITH-AES-256-CBC-SHA\ncipher AES-256-CBC\nauth SHA256\n\ndev tun\n# if dev tun doe"

hannesm:
This fixes #45

Also, cipher/ciphersuite and tls-version-min are applied to the TLS configuration (this was missing previously, so tls_version_min had no effect)

### DIFF
--- a/src/config.ml
+++ b/src/config.ml
@@ -39,6 +39,52 @@ let int_of_ping_interval = function
          violated"
   | `Seconds n -> n
 
+let (cipher_to_cs : string -> Tls.Ciphersuite.ciphersuite),
+    (cs_to_cipher : Tls.Ciphersuite.ciphersuite -> string) =
+  let map = [
+    "TLS-ECDHE-ECDSA-WITH-AES-256-GCM-SHA384", `ECDHE_ECDSA_WITH_AES_256_GCM_SHA384 ;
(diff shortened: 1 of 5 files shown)


reynir (approved):
Thanks!

We can (for another PR) consider the interaction between `--tls-version-min` and `--tls-cipher` and `--tls-ciphersuites`?

hannesm:
Indeed, we can now end up in nonsense configurations by specifying `tls-version-min 1.3` and only some `tls-cipher` (i.e. no `tls-ciphersuite`). Let's improve on that front once we find such configurations / uses.

reynir:
Reading the ocaml-tls code briefly it seems for tls 1.3 that tls-ciphers that are not ciphersuite13 are just not considered - I was worried perhaps an exception could occur. These nonsense configurations seem relatively benign (if useless).

hannesm:
IIRC the OCaml-TLS configuration constructor (Tls.Config.client/server) does some validity checks and raises an exception on useless configurations.

reynir:
Ah indeed `validate_common`  in  `config.ml` checks for this. I missed it at first read.

Issue #85, opened 2023-10-04
`--dev tunmir` not accepted


reynir:
I have a config file with `dev tunmir`, but the config doesn't accept this because `mir` is not a number in the range 0-128. Interestingly this works:

    dev mirtun
    dev-type tun

hannesm:
same as #43!?

reynir:
Yes!

PR #86, opened 2023-10-04
Config parser: allow dev tun<any string>, as in 'dev tunnel' or 'dev tunmir'

### LINKED ISSUE #43: config parser fails when name of tun device is tunnel
example here: 
https://github.com/whohoho/openvpn/commit/c37d7a1d80035413fdc3ed09fa78a7f1a7747028

hannesm:
Previously, only tun<numeric> (with numeric in range 0..128) was allowed, which lead to issues. In the end, the dev-type is `Tun if the name starts with 'tun', and `Tap if the name starts with 'tap' -- otherwise the 'dev-type' has to be specified.

Fixes #43 #85

### DIFF
--- a/src/config.ml
+++ b/src/config.ml
@@ -662,14 +662,18 @@ let a_dev =
   let a_device_name =
     choice
       [
-        ( string "tun"
-        *> option None
-             (a_number_range 0 128 >>| fun x -> Some ("tun" ^ string_of_int x))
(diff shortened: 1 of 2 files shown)


reynir (approved):

reynir:
This works for me. I'm not sure what the CI error is about, but it seems unrelated. Thanks!

PR #87, opened 2023-10-05
Fix Conf_map.pp_with_sep not using sep for client

reynir:
In `Config.client_generate_connect_options` we use comma as `sep`.


hannesm:
I fail to understand what this code is used for... but maybe a test case would be useful to avoid regressions?

reynir:
I think it is used for OCC. It is passed during key exchange. A log message in `Config.client_generate_connect_options` says "serialized connect options, probably incorrect: [...]" which is not so encouraging.

hannesm:
Thanks for your reply. I'm fine with merging this as is, since I read the code and there's no difference when looking at all use sites of generate_connect_options.

PR #88, opened 2023-10-05
Configuration: support "secret <filename>" and "secret [inline]"

### LINKED ISSUE #44: secret <file> is not parsed appropriately
openvpn_server_config_parser.exe static-home.conf 
openvpn_server_config_parser.exe: [INFO] Reading file "static-home.conf"
openvpn_server_config_parser.exe: [WARNING] IGNORING "up" "./home.up"
openvpn_server_config_parser.exe: [ERROR] : Error at byte offset 607: "secret static.key\n\n# Cipher to use\ncipher AES-256-CBC\n\n# OpenVPN 2.0 uses UDP port 1194 by default\n#"

hannesm:
Fixes #44

### DIFF
--- a/src/config.ml
+++ b/src/config.ml
@@ -892,6 +892,12 @@ let a_secret str =
   | Ok (a, b, c, d) -> Ok (B (Secret, (a, b, c, d)))
   | Error e -> Error e
 
+let a_secret_maybe_inline =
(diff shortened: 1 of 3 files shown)


reynir (approved):
I pushed a commit to change the names to align with the other inline payload parsers.

PR #89, opened 2023-10-05
Config: support (a) route-nopull (b) comp-lzo <bool> (c) verify-x509-name

### LINKED ISSUE #42: verify-x509-name is not parsed

hannesm:
Fixes #42

Only a subset of verify-x509-name is supported: type = "name" (not type = "subject", neither type = "name-prefix")

This allows to nearly parse wild-client.conf :)

### DIFF
--- a/src/config.ml
+++ b/src/config.ml
@@ -1951,8 +1951,8 @@ let merge_push_reply client (push_config : string) =
     | Dhcp_dns, _ -> true
     | Dhcp_ntp, _ -> true
     | Ifconfig, _ -> true
-    | Route, _ -> true
-    | Route_gateway, _ -> true
+    | Route, _ -> not (Conf_map.mem Route_nopull client)
+    | Route_gateway, _ -> not (Conf_map.mem Route_nopull client)
     | _ -> false
   in
   (* let naughty_server = in *)
--- a/src/config.ml
+++ b/src/config.ml
@@ -1937,8 +1937,8 @@ let merge_push_reply client (push_config : string) =
   let will_accept (type a) (k : a key) (v : a) =
     match (k, v) with
     (* whitelist keys we are willing to accept from server: *)
-    | Dhcp_disable_nbt, _ -> true
-    | Dhcp_domain, _ -> true
+    | Dhcp_disable_nbt, _ -> not (Conf_map.mem Route_nopull client)
+    | Dhcp_domain, _ -> not (Conf_map.mem Route_nopull client)
     | Mssfix, _ -> true
     | Tls_mode, `Client -> true
     | Tun_mtu, _ -> true
@@ -1948,8 +1948,8 @@ let merge_push_reply client (push_config : string) =
     | Ping_timeout, `Restart n when n >= 0 ->
         true (* TODO | Redirect_gateway, _ -> true *)
     (* TODO should verify IPs: *)
-    | Dhcp_dns, _ -> true
-    | Dhcp_ntp, _ -> true
+    | Dhcp_dns, _ -> not (Conf_map.mem Route_nopull client)
+    | Dhcp_ntp, _ -> not (Conf_map.mem Route_nopull client)
     | Ifconfig, _ -> true
     | Route, _ -> not (Conf_map.mem Route_nopull client)
     | Route_gateway, _ -> not (Conf_map.mem Route_nopull client)
--- a/src/config.ml
+++ b/src/config.ml
@@ -172,6 +172,7 @@ module Conf_map = struct
     | Route_delay : (int * int) k
     | Route_gateway : [ `Ip of Ipaddr.t | `Default | `Dhcp ] k
     | Route_metric : [ `Default | `Metric of int ] k
+    | Route_nopull : flag k
     | Script_security : int k
     | Secret : (Cstruct.t * Cstruct.t * Cstruct.t * Cstruct.t) k
     | Server : Ipaddr.V4.Prefix.t k
@@ -196,6 +197,7 @@ module Conf_map = struct
     | Tun_mtu : int k
     | Verb : int k
     | Verify_client_cert : [ `None | `Optional | `Required ] k
+    | Verify_x509_name : [`host] Domain_name.t k
 
   module K = struct
     type 'a t = 'a k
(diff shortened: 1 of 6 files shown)


src/miragevpn.mli
```diff
@@ -190,10 +192,10 @@ module Config : sig
     | Tls_ciphersuite : Tls.Ciphersuite.ciphersuite13 list k
 
     | Tls_crypt_v2_client : ((Cstruct.t * Cstruct.t * Cstruct.t * Cstruct.t) * Cstruct.t * bool) k
-    (** [Tls_crypt_v2_client (key, wkc, force_cookie) *)
+    (** [Tls_crypt_v2_client (key, wkc, force_cookie)] *)
```
reynir, line 195: Oops! Thanks.

src/engine.ml
```diff
@@ -448,8 +447,9 @@ let incoming_control_client config rng session channel now op data =
               Some (highest, highest)
             else
               Some (v, highest)
+        and peer_name = Config.find Verify_x509_name config
         in
-        Tls.(Engine.client (Config.client ?ciphers ?version ~certificates ~authenticator ()))
+        Tls.(Engine.client (Config.client ?ciphers ?version ?peer_name ~certificates ~authenticator ()))
```
reynir, line 452: The default type seems to be to check `CN`. I think this will check against the subjectAltName? See `--x509-username-field` in `man openvpn`. I think we would require `x509-username-field ext:subjectAltName` in order to do this.

I am also thinking if this is something we should rather pass to the authenticator. I don't remember what the ` ~peer_name` argument does.

src/config.ml
```diff
@@ -1938,8 +1951,8 @@ let merge_push_reply client (push_config : string) =
     | Dhcp_dns, _ -> true
     | Dhcp_ntp, _ -> true
     | Ifconfig, _ -> true
-    | Route, _ -> true
-    | Route_gateway, _ -> true
+    | Route, _ -> not (Conf_map.mem Route_nopull client)
+    | Route_gateway, _ -> not (Conf_map.mem Route_nopull client)
```
reynir, line 1955: I added these, but we need to add more. It's a bit unclear to me what exactly:

    When  used  with  --client  or  --pull, accept options pushed by
    server EXCEPT for routes,  block-outside-dns  and  dhcp  options
     like DNS servers.

src/engine.ml
hannesm: This is true, there's a discrepancy from OpenVPN and our implementation. But I think ours is more intuitive.

OpenVPN seems to only use the name of the subject of the certificate (esp. the Common Name fragment). We use the set of common name and SubjectAlternativeName.

I'm happy to document this behaviour somewhere. I'm also fine to have this discrepancy.

hannesm: The `~peer_name` specifies which hostname is expected on the server side. I.e. the `validate_chain` uses the peer_name for verifying the name in the X.509 certificate. It is also used in the ClientHello as SNI (Server Name Identification).

The "authenticator" is a function that gets passed the `?ip` and `?hostname` -- by the TLS stack.

src/config.ml
hannesm: yes, thanks. I looked into the openvpn code, and the route-nopull disables all "OPT_P_ROUTE | OPT_P_IPWIN32" things, namely:
- client-nat
- route
- route-ipv6
- route-metric
- redirect-gateway
- redirect-private
- block-ipv6
- foreign-option
- ip-win32
- dhcp-option
- tap-sleep
- dhcp-renew
- dhcp-pre-release
- register-dns
- block-outside-dns

(basically, a grep for OPT_P_ROUTE and OPT_P_IPWIN32 in options.c -- done on an old openvpn on my side).

I pushed a commit which completes your commit :)

src/engine.ml
hannesm: I documented that in [6bbb95a](https://github.com/robur-coop/miragevpn/pull/89/commits/6bbb95ae225afdef84b5cb43c21613cc88fe600f)

reynir: Hmm, ok. Yes, I think we should document it if we diverge.

I wrote this monstrosity if we want to support the openvpn behavior
```ocaml
        let authenticator =
          Option.fold (Config.find Verify_x509_name config)
            ~none:authenticator
            ~some:(fun peer_name ?ip ~host certs ->
                Result.bind
                  (authenticator ?ip ~host certs)
                  (fun r ->
                     match certs with
                     | [] -> Error `EmptyCertificateChain
                     | peer :: _ ->
                       let dn = X509.Certificate.subject peer in
                       if Option.equal String.equal (Some peer_name)
                           (X509.Distinguished_name.common_name dn)
                       then
                         Ok r
                       else
                         Error (`Msg "peer name mismatch")))
        in
```

reynir (approved):
Thank you for adding the note in the README! Looks good to me. ocaml-ci seems to be having a bad day :(

hannesm:
Thanks for your approval. I merged and opened the remaining things as separate issues.

src/engine.ml
reynir: We can invent a new `x509-username-field`, say `peer-name`, that does what we do and have that as a default, and then we can implement parsing `x509-username-field` -- and then users can specify `x509-username-field CN` if they really want that.
But yes - for another PR :)

PR #93, opened 2023-10-05
config: allow 'verb' to be present multiple times, the last one wins; support tls-version-max; fix tls-cipher/tls-ciphersuites

hannesm:


reynir (approved):
I find this tls-cipher/tls-ciphersuites split a bit weird. Some (or all?) tls 1.3 cipher suites are also tls 1.2 cipher suites, so when you specify --tls-cipher should you remove the 1.3 cipher suites that are 1.2 cipher suites as well that were not mentioned?! Can you say you want AES_128_CCM_SHA256 for 1.3 but not 1.2?! I am tempted to suggest we make tls-cipher{,suites} synonyms instead, but let's discuss that in another issue/pr.

hannesm:
The tls 1.3 and tls 1.2 ciphersuite names are actually disjoint. In OCaml-tls we have (the non-exported) Config.Ciphers.default13 and Config.Ciphers.default (which is default13 @ [... rest ...]). So, I'm fine with how this is dealt here -- though if we would expose more bindings from OCaml-TLS, we could specify less hardcoded values here :)

PR #97, opened 2023-10-05
use custom log sources for the library

hannesm:


hannesm:
this is just some janitor work, let's merge it :)

PR #98, opened 2023-10-05
packet.decode_header: reverse ack_message_id list

hannesm:
Previously, the order was not preserved, and since the order is relevant for the hmac computation, the hmac failed.

/cc @reynir

### DIFF
--- a/src/packet.ml
+++ b/src/packet.ml
@@ -84,14 +84,14 @@ let decode_header buf =
     (Cstruct.length buf >= hdr_len + (packet_id_len * arr_len) + rs)
     `Partial
   >>| fun () ->
-  let rec ack_message_id = function
-    | 0 -> []
+  let rec ack_message_id acc = function
+    | 0 -> acc
     | n ->
         let idx = pred n in
         let id = Cstruct.BE.get_uint32 buf (hdr_len + (packet_id_len * idx)) in
-        id :: ack_message_id idx
+        ack_message_id (id :: acc) idx
   in
-  let ack_message_ids = ack_message_id arr_len in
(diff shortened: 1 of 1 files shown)


hannesm:
the code is a bit intricate -- since we pass `arr_len` to the recursive function, we start to read the last acked message_id. Previously, we were then prepending that id to a list, and calling the recursive function, so we ended up with `<last>::<previous-to-last>::..::<first>::[]` -- which is reversed.

Since that function was not tail-recursive as well, I decided to switch it to a tail-recursive version where still the last message id is read first, but then the recursion is called with `<last-message-id>::[]`, and since it always prepends, we end up with the right order `<first>::..::<last>::[]`.

src/packet.ml
```diff
@@ -84,14 +84,14 @@ let decode_header buf =
...
+  let rec ack_message_id acc = function
+    | 0 -> acc
     | n ->
         let idx = pred n in
         let id = Cstruct.BE.get_uint32 buf (hdr_len + (packet_id_len * idx)) in
-        id :: ack_message_id idx
+        ack_message_id (id :: acc) idx
   in
-  let ack_message_ids = ack_message_id arr_len in
+  let ack_message_ids = ack_message_id [] arr_len in
```
reynir, line 94: GitHub code review doesn't let me select all the relevant bits for some odd reason. How about this:
```suggestion
  let ack_message_id idx =
    Cstruct.BE.get_uint32 buf (hdr_len + (packet_id_len * idx))
  in
  let ack_message_ids = List.init arr_len ack_message_id in
```

PR #100, opened 2023-10-06
miragevpn-client-lwt: use connect(2) for udp

### LINKED ISSUE #96: miragevpn-client-lwt privileged port / running as a regular user
The following code generates a random port, but if you're unlucky it may be a privileged port. This may cause the application to fail due to that.
https://github.com/robur-coop/miragevpn/blob/b816eb071622a67b9adb72f8d5a0eeaba228cb14/app/miragevpn_client_lwt.ml#L227

It may not be an issue because I think other parts require privileges (though maybe not as much?). Perhaps we should check for uid = 0.

reynir:
Probably we can share more code with tcp now. Fixes #96.

### DIFF
--- a/app/miragevpn_client_lwt.ml
+++ b/app/miragevpn_client_lwt.ml
@@ -82,11 +82,11 @@ let write_multiple_to_fd fd bufs =
       else Lwt.return r)
     true bufs
 
-let write_udp sa fd data =
+let write_udp fd data =
   Lwt.catch
     (fun () ->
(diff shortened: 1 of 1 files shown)


hannesm:
LGTM. I was not aware of the fact that `connect(2)` works fine on SOCK_DGRAM -- but it does :)

And indeed, more code could be shared. Also, could you run `dune bu @fmt --auto` to conform to the code formatting? Thanks a lot!

PR #101, opened 2023-10-08
parse and validate server config

hannesm:
This introduces a new function `parse_server` which parses and validates a server configuration. Also, `is_client_config` is introduced, which checks whether `Remote` is present or not.

This superseeds #50 #38

### DIFF
--- a/mirage-server/unikernel.ml
+++ b/mirage-server/unikernel.ml
@@ -15,7 +15,7 @@ struct
     | Error e -> Error (`Msg (Fmt.to_to_string FS.pp_error e))
     | Ok data ->
         let string_of_file _ = Error (`Msg "no string_of_file support") in
-        Miragevpn.Config.parse ~string_of_file data
+        Miragevpn.Config.parse_server ~string_of_file data
 
   let start _ _ _ _ s data =
     read_config data >>= function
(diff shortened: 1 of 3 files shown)


reynir:
Thanks! I wonder how you are supposed to use `is_client_config`: We have `Config.parse` which does not try to validate the client- or server specific details. But then if we discover it is a client config (server config) we don't get the client (server) defaults merged. Maybe we need to expose some more functions?

hannesm:
> Thanks! I wonder how you are supposed to use `is_client_config`: We have `Config.parse` which does not try to validate the client- or server specific details. But then if we discover it is a client config (server config) we don't get the client (server) defaults merged. Maybe we need to expose some more functions?

Yes, I guess we will need some more functionality (and figure out the exact use case for it). So, as a first step I removed the `is_client_config` -- we can easily re-add it at a later point when we're clear what we need.

hannesm:
CI failure is due to the fact that cirrus tests all the unikernels -- which each pins miragevpn to the main branch of the repository -- and here the Miragevpn.Config.parse_server is not yet bound.

We should adapt the CI scripts so that the miragevpn of the specific branch / checkout is used. But that's something that can be addressed in a separate PR -- I opened issue #106 for tracking this.

reynir (approved):

PR #105, opened 2023-10-09
Tls crypt v2

reynir:
Work in progress implementation of tls-crypt-v2 with @dinosaure.

TODO:

- [ ] SHA256 hmac
- [ ] extending the state with tls-crypt-v2 keys

### DIFF
--- a/src/engine.ml
+++ b/src/engine.ml
@@ -478,7 +478,7 @@ let maybe_push_reply config = function
 
 let incoming_control_client config rng session channel now op data =
   match (channel.channel_st, op) with
-  | Expect_reset, (Packet.Hard_reset_server | Packet.Soft_reset) ->
+  | Expect_reset, (Packet.Hard_reset_server_v2 | Packet.Soft_reset_v2) ->
       (* for rekey we receive a soft_reset -- a bit alien that we don't send soft_reset *)
       (* we reply with ACK + TLS client hello! *)
       let tls, ch =
@@ -598,7 +598,7 @@ let init_channel how session hmac_algorithm keyid now ts =
 let incoming_control_server is_not_taken config rng session channel _now _ts
     _key op data =
   match (channel.channel_st, op) with
-  | Expect_reset, (Packet.Hard_reset_client | Packet.Soft_reset) ->
+  | Expect_reset, (Packet.Hard_reset_client_v2 | Packet.Soft_reset_v2) ->
       (* TODO may need to do client certificate authentication here! *)
       let _ca, server, key =
         ( Config.get Ca config,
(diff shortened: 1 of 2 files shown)


reynir:
Rebased on main, and added a commit that implements `Hard_reset_client_v3` for encoding anyway.

The server prints the following:
```
2023-10-13 14:00:46 us=998471 192.168.1.84:39138 tag      : d812c0ea 69ea3fd9 5734e8ca eabaedef 71bef6be 1114362b 00000001 6529316e
2023-10-13 14:00:46 us=998482 192.168.1.84:39138 tag_check: c25c5318 8556d57e 1375ff0a 37f3335f 470cb9c6 70ccd54c bec8f265 bab9e9ce
```
`00000001 6529316e` looks like the metadata -- a timestamp 1697198446 or Fri 13 Oct 2023 12:00:46 UTC -- so likely something is not aligned correctly or missing.

reynir:
Ah no that timestamp is very recent. It must be the timestamp from the header.

reynir:
I discovered that the replay packet id (packet id + timestamp) and the hmac are swapped in tls-crypt packages(!) Furthermore, the hmac seems to be computed differently as well!

reynir:
An important question I didn't investigate before now is which keys are used in what direction.
```C
void
tls_crypt_init_key(struct key_ctx_bi *key, struct key2 *keydata,
                   const char *key_file, bool key_inline, bool tls_server)
{
    const int key_direction = tls_server ?
                              KEY_DIRECTION_NORMAL : KEY_DIRECTION_INVERSE;
    struct key_type kt = tls_crypt_kt();
    if (!kt.cipher || !kt.digest)
    {
        msg(M_FATAL, "ERROR: --tls-crypt not supported");
    }
    crypto_read_openvpn_key(&kt, key, key_file, key_inline, key_direction,
                            "Control Channel Encryption", "tls-crypt", keydata);
}
```
From the above code a server will use `KEY_DIRECTION_NORMAL` which corresponds to `--key-direction 0` or what we call `` `Outgoing`` in our code. The client will use `KEY_DIRECTION_INVERSE (aka. `--key-direction 1` or `` `Incoming``).

reynir:
Despite what I initially thought the `--auth hmac` does not apply when using tls-crypt[-v2]. It is always sha256 not just for the wrapped key but as well for the control message auth tag:
https://github.com/OpenVPN/openvpn/blob/master/src/openvpn/tls_crypt.h#L60
https://github.com/OpenVPN/openvpn/blob/master/src/openvpn/tls_crypt.h#L89

reynir:
Since the auth tag offset and encrypted payload offset are statically known we can move encryption/decryption and hmac computation back to engine.ml. Furthermore, it may make sense to split the functions in packet.ml into tls-auth and tls-crypt as they are different formats (if ever so slightly different).

hannesm:
superseeded by #127

PR #107, opened 2023-10-11
Add a little tool to decode tls-crypt-v2 keys

dinosaure:
This little code show how to decode PEM format keys generated by `openvpn` (see `openvpn --genkey --tls-crypt-v2`). They show metadata and check if the given key is valid. Assume that you have `server.pem` (the server key) and `Alice.pem` (the client key), you can use this little tool with:
```shell-session
$ dune exec app/tls_crypt_v2.exe -- tls-crypt-v2 server server.pem
$ dune exec app/tls_crypt_v2.exe -- tls-crypt-v2 client Alice.pem server.pem
```

The executable requires `hxd` which is probably not installed and required by the OPAM file.

### DIFF
--- a/app/tls_crypt_v2.ml
+++ b/app/tls_crypt_v2.ml
@@ -1,0 +1,206 @@
+let error_msgf fmt = Fmt.kstr (fun msg -> Error (`Msg msg)) fmt
+
+module Tls_crypt_v2_key : sig
+  type t
+
+  val of_cstruct : Cstruct.t -> (t, [> `Msg of string ]) result
+  val to_string : t -> string
+  val cipher_key : t -> Mirage_crypto.Cipher_block.AES.CTR.key
+  val hmac : t -> Cstruct.t
+  val equal : t -> t -> bool
+  val pp_hum : t Fmt.t
+end = struct
+  type t = Cstruct.t (* cipher (64 bytes) + hmac (64 bytes) *)
+
+  let of_cstruct cs =
+    if Cstruct.length cs <> 128
+    then error_msgf "Invalid tls-crypt-v2 key"
+    else
+      let len = Array.fold_left max 0 Mirage_crypto.Cipher_block.AES.CTR.key_sizes in
+      try
+        let secret = Cstruct.sub cs 0 len in
+        let _ = Mirage_crypto.Cipher_block.AES.CTR.of_secret secret in Ok cs
+      with exn -> error_msgf "Invalid AES-CTR secret key: %S" (Printexc.to_string exn)
+
+  let to_string t = Cstruct.to_string t
+
+  let cipher_key cs =
+    let len = Array.fold_left max 0 Mirage_crypto.Cipher_block.AES.CTR.key_sizes in
+    Mirage_crypto.Cipher_block.AES.CTR.of_secret (Cstruct.sub cs 0 len)
+
+  let hmac cs = Cstruct.sub cs 64 64
+  let equal a b = Eqaf_cstruct.equal a b
+
+  let pp_hum ppf cs =
+    let cipher_key = Cstruct.(to_string (sub cs 0 64)) in
+    let hmac = Cstruct.(to_string (sub cs 64 64)) in
+    Fmt.pf ppf "Cipher Key: @[<hov>%a@]\n%!" (Hxd_string.pp Hxd.default) cipher_key;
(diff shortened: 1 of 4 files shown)


app/tls_crypt_v2.ml
```diff
@@ -28,7 +28,7 @@ end = struct
     let len = Array.fold_left max 0 Mirage_crypto.Cipher_block.AES.CTR.key_sizes in
     Mirage_crypto.Cipher_block.AES.CTR.of_secret (Cstruct.sub cs 0 len)
 
-  let hmac cs = Cstruct.sub cs 64 64
```
dinosaure: Are you sure about that? In `openvpn`, we have this `define`:
```c
#define MAX_HMAC_KEY_LENGTH 64
```

reynir: Yes, it's described here: https://github.com/OpenVPN/openvpn/blob/master/doc/tls-crypt-v2.txt#L65 (256 bits = 32 bytes)

Besides, we have this `#define` even though we use only the first 32 bytes of the encryption key:
```C
#define MAX_CIPHER_KEY_LENGTH 64
```

(IMO it's a bit messy to have these keys that we only use half of)

reynir: Another argument for is that it actually authenticates the wrapped key now that we check the tag and computed tag are equal (and not the opposite).

dinosaure: Hmmmhmm, ok we must clarify that because we got so trouble when we read the `openvpn` code.

reynir: Yes, this needs to be documented. It seems it's a thing throughout the openvpn code to only use the first N bytes of the keys where N depends on the cipher/hmac. See https://github.com/OpenVPN/openvpn/blob/master/src/openvpn/crypto_openssl.c#L1226-L1241

reynir: I tried to document it here: https://git.robur.coop/robur/openvpn-spec/pulls/7

app/tls_crypt_v2.ml
```diff
@@ -133,6 +158,65 @@ let load_tls_crypt_v2_server_key filename =
...
+  let a = Tls_crypt_v2_key.generate ?g () in
+  let b = Tls_crypt_v2_key.generate ?g () in
+  (a, b, metadata)
+
+let tls_crypt_v2_wrap_client server_key (a, b, metadata) =
+  let a = Tls_crypt_v2_key.to_cstruct a in
+  let b = Tls_crypt_v2_key.to_cstruct b in
+  let cs = Cstruct.concat [ a; b ] in
+  let metadata = Metadata.to_cstruct metadata in
+  let net_len = (128 * 2) + (Cstruct.length metadata) + _TLS_CRYPT_V2_TAG_SIZE + 2 in
```
reynir, line 189: Not necessarily clearer:
```suggestion
  let net_len = Cstruct.length a + Cstruct.length b + Cstruct.length metadata + Mirage_crypto.Hash.SHA256.digest_size + 2 in
```

reynir: Or since we bind `cs := a||b`
```suggestion
  let net_len = Cstruct.length cs + Cstruct.length metadata + Mirage_crypto.Hash.SHA256.digest_size + 2 in
```

app/tls_crypt_v2.ml
```diff
@@ -133,6 +158,65 @@ let load_tls_crypt_v2_server_key filename =
...
+  let kc =
+    let (a, b, _) = client_key in
+    Cstruct.concat [ Tls_crypt_v2_key.to_cstruct a; Tls_crypt_v2_key.to_cstruct b ] in
+  let wkc = tls_crypt_v2_wrap_client server_key client_key in
+  let payload = Cstruct.concat [ kc; wkc ] in
+  let payload = Cstruct.to_string payload in
+  let b64 = Base64.encode_string ~pad:true payload in
+  let oc = open_out filename in
+  line oc "-----BEGIN OpenVPN tls-crypt-v2 client key-----";
+  let lines = String.split_at 65 b64 in
```
reynir, line 215: 64?
```suggestion
  let lines = String.split_at 64 b64 in
```

hannesm:
I marked this as ready for review, and suggest to review and merge it sooner than later :)

hannesm:
@reynir @dinosaure WDYT? should we merge (when CI passes)?

reynir:
I think this is good to merge.

What we can consider doing (in another PR) is:
- Use cmdliner,
- Output static keys (for `--secret`, `--tls-auth`),
- share more code with the library.

Issue #82, opened 2023-10-04
config --auth not utilized


hannesm:
The --auth can switch the HMAC algorithm, default is SHA1 (which atm is hardcoded).

hannesm:
from the manual page:
```
              Authenticate data channel packets and (if enabled) tls-auth
              control channel packets with HMAC using message digest algorithm
              alg. (The default is SHA1 ). HMAC is a commonly used message
              authentication algorithm (MAC) that uses a data string, a secure
              hash algorithm and a key to produce a digital signature.

              The OpenVPN data channel protocol uses encrypt-then-mac (i.e.
              first encrypt a packet then HMAC the resulting ciphertext),
              which prevents padding oracle attacks.

              If an AEAD cipher mode (e.g. GCM) is chosen then the specified
              --auth algorithm is ignored for the data channel and the
              authentication method of the AEAD cipher is used instead. Note
              that alg still specifies the digest used for tls-auth.

              In static-key encryption mode, the HMAC key is included in the
              key file generated by --genkey. In TLS mode, the HMAC key is
              dynamically generated and shared between peers via the TLS
              control channel. If OpenVPN receives a packet with a bad HMAC it
              will drop the packet. HMAC usually adds 16 or 20 bytes per
              packet. Set alg=none to disable authentication.
```

What we do: we just use SHA1, unconditionally. Thus I'd expect any AEAD tls-cipher to fail. Also, setting alg=none (to disable authentication) is not supported (as far as I read the code).

reynir:
I think `--tls-cipher` and `--auth` are orthogonal. The former is for the tls handshake while the latter is for the session keys.

hannesm:
#108 solves this

PR #108, opened 2023-10-11
Support `auth`

reynir:
A (nice-ish) parse error is raised if the hash is not one of those in mirage-crypto. Furthermore, `is_valid_config` checks that only `auth sha1` may be present for now.

### DIFF
--- a/src/config.ml
+++ b/src/config.ml
@@ -123,6 +123,7 @@ module Conf_map = struct
          - Remove it from Ignored Directives and/or Unimplemented Directives
   *)
   type 'a k =
+    | Auth : Mirage_crypto.Hash.hash k
     | Auth_nocache : flag k
     | Auth_retry : [ `Interact | `Nointeract | `None ] k
     | Auth_user_pass : (string * string) k
@@ -240,9 +241,13 @@ module Conf_map = struct
       (fun err -> `Msg ("not a valid config: " ^ err))
       ( ensure_mem Cipher "config must specify 'cipher AES-256-CBC'"
       >>= fun () ->
-        if mem Cipher t && get Cipher t <> "AES-256-CBC" then
-          Error "currently only supported Cipher is 'AES-256-CBC'"
-        else Ok () )
+        (if mem Cipher t && get Cipher t <> "AES-256-CBC" then
+           Error "currently only supported Cipher is 'AES-256-CBC'"
+         else Ok ())
+        >>= fun () ->
+        match find Auth t with
+        | None | Some `SHA1 -> Ok ()
+        | Some _h -> Error "currently only supported auth is 'SHA1'" )
 
   let is_valid_server_config t =
     let ensure_mem k err = if mem k t then Ok () else Error err in
@@ -379,7 +384,18 @@ module Conf_map = struct
         | `TLS_1_1 -> "1.1"
         | `TLS_1_0 -> "1.0")
     in
+    let pp_digest_algorithm ppf h =
+      Fmt.string ppf
+        (match h with
+        | `MD5 -> "MD5"
+        | `SHA1 -> "SHA1"
+        | `SHA224 -> "SHA224"
+        | `SHA256 -> "SHA256"
+        | `SHA384 -> "SHA384"
+        | `SHA512 -> "SHA512")
+    in
     match (k, v) with
+    | Auth, h -> p () "auth %a" pp_digest_algorithm h
     | Auth_nocache, () -> p () "auth-nocache"
     | Auth_retry, `None -> p () "auth-retry none"
     | Auth_retry, `Nointeract -> p () "auth-retry nointeract"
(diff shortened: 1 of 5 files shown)


src/config.ml
```diff
@@ -240,9 +241,13 @@ module Conf_map = struct
...
-        if mem Cipher t && get Cipher t <> "AES-256-CBC" then
-          Error "currently only supported Cipher is 'AES-256-CBC'"
-        else Ok () )
+        (if mem Cipher t && get Cipher t <> "AES-256-CBC" then
+           Error "currently only supported Cipher is 'AES-256-CBC'"
+         else Ok ())
+        >>= fun () ->
+        match find Auth t with
+        | None | Some `SHA1 -> Ok ()
+        | Some _h -> Error "currently only supported auth is 'SHA1'" )
```
hannesm, line 250: to me it looks like you've adjusted all the parts where `SHA1` was hardcoded... and my current guess it, any other (supported) hmac just works. did you try that?

src/config.ml
```diff
@@ -379,7 +384,18 @@ module Conf_map = struct
...
     in
+    let pp_digest_algorithm ppf h =
+      Fmt.string ppf
+        (match h with
+        | `MD5 -> "MD5"
+        | `SHA1 -> "SHA1"
+        | `SHA224 -> "SHA224"
+        | `SHA256 -> "SHA256"
+        | `SHA384 -> "SHA384"
+        | `SHA512 -> "SHA512")
```
hannesm, line 395: if we move this somwhere, we may reuse it in `State.pp`

src/engine.ml
```diff
@@ -101,20 +105,28 @@ let hmacs config =
   match Config.find Tls_auth config with
   | None -> Error (`Msg "no tls auth payload in config")
   | Some (direction, _, hmac1, _, hmac2) ->
+      let hmac_len =
+        Option.value ~default:`SHA1 (Config.find Auth config)
+        |> Mirage_crypto.Hash.digest_size
+      in
```
hannesm, line 111: I'd propose to move ``Auth `SHA1`` into the default configuration, and avoid the `Option.value ~default` (that appears thrice in this PR). WDYT?

src/state.ml
```diff
@@ -192,8 +193,9 @@ type session = {
 }
 
 let init_session ~my_session_id ?(their_session_id = 0L) ?(compress = false)
-    ?(protocol = `Tcp) ~my_hmac ~their_hmac () =
+    ?(protocol = `Tcp) ?(hmac_algorithm = `SHA1) ~my_hmac ~their_hmac () =
```
hannesm, line 196: I wonder, is there negotiation of the `auth`? I.e. if a client has one in its configuration, can the server deselect it and choose a different one (i.e. for PUSH_NOTIFY)?

If the answer is no, why should we keep it in `type session`? We can just use the `Config.get Auth` whenever needed.

src/engine.ml
reynir: Yea, that makes sense. That's what openvpn does too in fact!

hannesm:
fixes #82

src/state.ml
reynir: I find the story in openvpn about cipher (and hmac?) negotiation hard to follow. At the moment I don't remember any negotiation about `auth` - except if the cipher is AEAD where `auth` is ignored.

I added it to `session` because we seem to push that "further" than `config`. I also had in mind that we need to choose SHA256 as `auth` for control packets when using tls-crypt-v2, but I'm not sure how I got this idea and I don't think it is the case.

src/config.ml
reynir: I have not tried that yet. With main branch I get errors on the openvpn 2.6 server such as `TLS Error: unknown opcode received from [AF_INET]192.168.1.84:58614 op=12` and a few of `TLS Error: client->client or server->server connection attempted from [AF_INET]192.168.1.84:58614`. It seems to get stuck after the handshake. I should test this because if I can get as far with a different `auth` we exercise the code path and it's no worse.

hannesm: oh, even after #98 you're not able to establish a connection?

reynir: With the other issues fixed I tried with `auth sha256`, and that does not seem to work: `TLS Error: cannot locate HMAC in incoming packet from [AF_INET]192.168.1.84:37694`

src/state.ml
reynir: I have refactored the code to not store the hmac algorithm in `type session`.

I have not seen anything that suggests the hmac algorithm can be negotiated - other than if you use an AEAD cipher. We probably need to restructure the code in order to support AEAD. I will not think about that right now though.

reynir:
Now rebased on main and ready for review / merge.

src/config.ml
reynir: It was due to an optional argument with default value `` `SHA1`` that was never passed.

PR #109, opened 2023-10-12
Allow any TLS session tickets etc. to arrive in the TLS_established state

hannesm:
Just continue, and once data arrived continue with the KDF

### DIFF
--- a/src/engine.ml
+++ b/src/engine.ml
@@ -521,31 +521,36 @@ let incoming_control_client config rng session channel now op data =
   | TLS_established (tls, key), Packet.Control -> (
       let open Result.Infix in
       incoming_tls tls data >>= fun (tls', tls_resp, d) ->
-      maybe_kdf session key d >>= fun (tls_data, keys) ->
-      let config = merge_server_config_into_client config tls_data in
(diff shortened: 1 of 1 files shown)


hannesm:
it has been confirmed by reynir that this is a good fix.

reynir:
Do we need to apply something similar for the server? I guess maybe no, but I don't know TLS all that well.

hannesm:
the purpose of a SessionTicket is to avoid the (computationally expensive) asymmetric crypto handshake in subsequent sessions. Now, with TLS 1.3 the SessionTicket stuff changed a bit, and the server now sends us one (or many) SessionTickets (we should store and restore them in the future).

When we are the server, we should also generate SessionTickets and allow them to used. But this is something we can put in a separate issue -- since there are more important issues to tackle. And in the end, SessionTicket aren't that worthy for normal PCs (esp. with elliptic curves, the asymmetric operations aren't that expensive anymore). There's as well the issue of storing the tickets and restoring them, plus expiry / lifetime. OCaml-TLS has an API for SessionTicket, but I've not seen anyone using it...

PR #111, opened 2023-10-13
Add length prefix to static ping messages for TCP

reynir:
It seems ping messages were not prefixed with message length in TCP when running as a static client.

@hannesm can you try this fix and see if it solves the issue you observed?


hannesm:
excellent, this works great!

I force-pushed with ocamlformatted sources, and will merge! :)

PR #113, opened 2023-10-13
app: report log with a prefixed timestamp

hannesm:


reynir (approved):
YES PLEASE!

hannesm:
on the downside, we loose colours in the log output. and I'm also curious whether we're interested in relative or absolute time (and precision)... any opinion? I'm happy to merge this as-is, and maybe a "logs reporter library" is something else to do (and/or convert mirage-logs to become one that we're satisfied with)

PR #115, opened 2023-10-14
re-use the Logs.format_reporter (to print colors and remove code)

hannesm:


reynir:
Thanks!

Does this print the log source? IIRC from my experiments it didn't print the log source, but my code was slightly different. It would be very nice to have the log source.

reynir (approved):
... On the other hand it is very nice to reduce the amount of code.

hannesm:
I adjusted this to (a) print log source (b) print tags (c) use colors if desired

PR #117, opened 2023-10-14
run tests and crowbar tests in separate binaries

hannesm:
previously, there was a single test binary which exit code was always 0 (thus test failures weren't detected in CI)

- fix tests: default configuration now has auth SHA1
- enable wild-client and windows-riseup-client configurations
- for the latter, ignore block-outside-dns and remote-cert-eku


reynir:
Thanks!

There is one CI job, debian on 32 bit x86, that fails. The failure seems to be in unrelated(?!) packages.

PR #118, opened 2023-10-16
Implement key direction for --secret

### LINKED ISSUE #116: static key mode only partially implemented
esp. no direction, i.e. the secret is always bi-directionally using the first key (see `let secret = ... ` in engine.ml).

reynir:
Fixes #116.

This needs to be tested, and there is some room for refactorings.


hannesm:
LGTM

PR #119, opened 2023-10-16
Handle the AUTH_FAILED after a PULL_REQUEST

dinosaure:
This commit clean-up the OCaml client to terminate correctly if we got such error.

### DIFF
--- a/src/engine.ml
+++ b/src/engine.ml
@@ -463,6 +463,12 @@ let maybe_push_reply config = function
         if String.starts_with str ~prefix:p_r then
           let opts = String.sub str p_r_len (String.length str - p_r_len) in
           Config.merge_push_reply config opts
+        else if String.starts_with str ~prefix:"AUTH_FAILED" then
+          let msg =
+            "Authentication failed: "
+            ^ (List.tl (String.split_on_char ',' str) |> String.concat ",")
+          in
+          Error (`Msg msg)
         else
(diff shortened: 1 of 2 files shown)


src/engine.ml
```diff
@@ -463,6 +463,9 @@ let maybe_push_reply config = function
         if String.starts_with str ~prefix:p_r then
           let opts = String.sub str p_r_len (String.length str - p_r_len) in
           Config.merge_push_reply config opts
+        else if String.starts_with str ~prefix:"AUTH_FAILED" then
+          let msg = List.tl (String.split_on_char ',' str) |> String.concat "," in
+          Error (`Msg msg)
```
hannesm, line 468: you test for "AUTH_FAILED" prefix, and then do a split on `,` and a List.tl -- would you mind to test for "AUTH_FAILED,"? (so we ensure there's at least two elements after the split?

hannesm: or otherwise do a `String.sub str 11 (String.length str - 11)` -- for AUTH_FAILED being 11 chars... (or use "AUTH_FAILED," and 12)...

dinosaure: We probably can prepend the message with something like `authentication failed`, at least, the `List.tl` should never failed where `String.split_on_char` can not return an empty list. With the current code, the worst case will be to have an empty message.

dinosaure: Done 👍

hannesm:
thanks

Issue #102, opened 2023-10-09
peer-info and cipher negotiation


reynir:
From browsing the openvpn source code and documentation I discovered the "peer-info" mechanism. It is used to push information about the client such as version, platform, supported data channel ciphers: `IV_VER`, `IV_PLAT`, `IV_CIPHERS`. The capability to negotiate data channel ciphers is communicated by pushing `IV_NCP=2`, and there is a quirk with how openvpn 2.4 handles that (TODO: describe the quirk), but otherwise I believe it is the first value in the server's `IV_CIPHERS` the client also announced.

The peer-info is pushed in tls-data, and the code has a comment that additional data is pushed - this is the peer-info. It seems an openvpn server will push the empty wire string if the client didn't send any peer-info.
https://github.com/robur-coop/miragevpn/blob/64f2b7dd40613d935487c526d3089244ddd4816b/src/packet.ml#L375-L381

hannesm:
Thanks for your investigation. Seems like this protocol is growing in complexity... :/ I also noticed there's --data-cipher now... I'm not sure how much of that stuff we actually want to support (and would focus on the other things, such as tls-crypt-v2, first) :)

hannesm:
fixed by #121

Issue #99, opened 2023-10-06
Config.is_valid_client_config needs additional checks for Client_static


reynir:
If I remember correctly the engine will query for `Ifconfig`, `Route_gateway` and maybe more. Maybe it is a bad config, but I was missing `route-gateway` (I did not intend to route the traffic outside the virtual private network).

hannesm:
Aren't some of these options only later transported (by the server) via the PUSH_NOTIFY?

reynir:
Hmm, true. These are usually pushed by the server (but can be configured by the client). Maybe the way forward would be to not expect `Ifconfig` and `Route_gateway` to be present, and maybe the latter only when the former is present.

Issue #36, opened 2020-03-28
Does not support connecting to server that does not use tls-auth


whohoho:
Its not possible to connect to a server that does not use tls-auth

hannesm:
this is correct. if you're looking for static pre-shared keys mode, please take a look at #11 (which I'm working on, re-started my efforts yesterday evening).

hannesm:
or is there some other authentication mode you'd like to have support for? if it is static pre-shared keys (`secret` in openvpn configuration), I think this is a duplicate of #11

whohoho:
no, i am talking about the optional HMAC signature.

When a server does not have that enabled, and it is enabled on the client, the client cannot connect.

https://openvpn.net/community-resources/hardening-openvpn-security/
The tls-auth HMAC signature provides an additional level of security above and beyond that provided by SSL/TLS. It can protect against: DoS attacks or port flooding on the OpenVPN UDP port. Port scanning to determine which server UDP ports are in a listening state.


packet format here: https://build.openvpn.net/doxygen/network_protocol.html

hannesm:
ok, and do you actively use such a configuration?

whohoho:
I don't actively use such a configuration but its quite common to have openvpn configured that way:

Here is 1 example, checking some other ones to see how they are configured:
https://riseup.net/vpn/vpn-red/windows/windows-riseup.ovpn

whohoho:
another example: https://github.com/leapcode/bitmask-dev/blob/master/src/leap/bitmask/vpn/helpers/linux/bitmask-root#L133

cfcs:
I also think there's value in skipping the computational + bandwidth overhead for an essentially useless HMAC with a global key shared amongst all users. This used to be useful for two things:
- If you trusted all clients, this could be used to limit exposure to `openssl`
- If you had clients behind very simplistic DPI boxes this would make the traffic look less like TLS, but TBH I am not convinced that such boxes exist (they could just check for the x509 certs).

With no `openssl`, and no examples of DPI boxes that are actually fooled by this, I'd be in favor of not having this be mandatory.

hannesm:
see #291, thanks for the report. sorry it took us so long to implement.

PR #121, opened 2023-10-17
support peer-info and NCP in client

reynir:
This PR is not meant to be merged as-is. It serves to share some initial work towards #102.

### DIFF
--- a/src/config.ml
+++ b/src/config.ml
@@ -152,7 +152,7 @@ module Conf_map = struct
     | Connect_retry : (int * int) k
     | Connect_retry_max : [ `Unlimited | `Times of int ] k
     | Connect_timeout : int k
-    | Data_ciphers : [ `AES_128_GCM | `AES_256_GCM | `CHACHA20_POLY1305 ] list k
+    | Data_ciphers : aead_cipher list k
     | Dev : ([ `Tun | `Tap ] * string option) k
     | Dhcp_disable_nbt : flag k
     | Dhcp_dns : Ipaddr.t list k
--- a/src/config.ml
+++ b/src/config.ml
@@ -2131,11 +2131,12 @@ let client_generate_connect_options t =
         | B (Link_mtu, _) -> true
         | B (Pull, _) -> true
         | B (Tls_mode, `Client) -> true
+        | B (Auth, _) -> true
         | _ -> false)
       t
   in
   let serialized =
-    Fmt.str "V4,tls-client,auth SHA1,keysize 256,%s%a"
+    Fmt.str "V4,tls-client,%s%a"
       (match Conf_map.find Dev t with
       | Some (`Tun, _) -> "dev-type tun,"
       | Some (`Tap, _) -> "dev-type tap,"
--- a/src/config.ml
+++ b/src/config.ml
@@ -2166,7 +2166,7 @@ let client_generate_connect_options t =
   let excerpt =
     Conf_map.filter
       (function
-        | B (Cipher, _) -> false (* TODO: only relevant if no NCP is used *)
+        | B (Cipher, _) -> not (Conf_map.mem Tls_mode t)
         | B (Comp_lzo, _) -> true
         | B (Tun_mtu, _) -> true
         | B (Link_mtu, _) -> true
--- a/src/config.ml
+++ b/src/config.ml
@@ -1333,18 +1333,18 @@ let a_server =
   | Ok cidr -> return (`Entry (B (Server, cidr)))
   | Error (`Msg m) -> fail m
 
-let aead_cipher c =
+let aead_cipher ~ctx c =
   match String.uppercase_ascii c with
   | "AES-128-GCM" -> return `AES_128_GCM
   | "AES-256-GCM" -> return `AES_256_GCM
   | "CHACHA20-POLY1305" -> return `CHACHA20_POLY1305
-  | _ -> Fmt.kstr fail "Unknown cipher %S" c
+  | _ -> Fmt.kstr fail "Unknown or unsupported cipher for %S: %S" ctx c
 
 let a_cipher =
   string "cipher" *> a_whitespace *> a_single_param >>= fun v ->
   (match String.uppercase_ascii v with
   | "AES-256-CBC" -> return `AES_256_CBC
-  | c -> aead_cipher c)
+  | c -> aead_cipher ~ctx:"cipher" c)
   >>| fun v -> `Entry (B (Cipher, v))
 
 let a_data_ciphers =
@@ -1353,7 +1353,7 @@ let a_data_ciphers =
   List.fold_left
     (fun acc c ->
       acc >>= fun acc ->
-      aead_cipher c >>| fun c -> c :: acc)
+      aead_cipher ~ctx:"data-ciphers" c >>| fun c -> c :: acc)
     (return []) ciphers
   >>| fun ciphers ->
   let ciphers = List.rev ciphers in
--- a/src/config.ml
+++ b/src/config.ml
@@ -102,13 +102,17 @@ let ( (cipher13_to_cs13 : string -> Tls.Ciphersuite.ciphersuite13),
       | Some c -> c),
     fun c -> List.assoc c rev_map )
 
-type cipher = [ `AES_256_CBC | `AES_128_GCM | `AES_256_GCM | `CHACHA20_POLY1305 ]
+type aead_cipher = [ `AES_128_GCM | `AES_256_GCM | `CHACHA20_POLY1305 ]
+type cipher = [ aead_cipher | `AES_256_CBC ]
 
-let cipher_to_string : cipher -> string = function
-  | `AES_256_CBC -> "AES-256-CBC"
+let aead_cipher_to_string : aead_cipher -> string = function
   | `AES_128_GCM -> "AES-128-GCM"
   | `AES_256_GCM -> "AES-256-GCM"
   | `CHACHA20_POLY1305 -> "CHACHA20-POLY1305"
+
+let cipher_to_string : cipher -> string = function
+  | #aead_cipher as c -> aead_cipher_to_string c
+  | `AES_256_CBC -> "AES-256-CBC"
 
 module Conf_map = struct
   type flag = unit
@@ -148,10 +152,7 @@ module Conf_map = struct
     | Connect_retry : (int * int) k
     | Connect_retry_max : [ `Unlimited | `Times of int ] k
     | Connect_timeout : int k
-    | Data_ciphers
-        : [ `AES_256_CBC | `AES_128_GCM | `AES_256_GCM | `CHACHA20_POLY1305 ]
-          list
-          k
+    | Data_ciphers : [ `AES_128_GCM | `AES_256_GCM | `CHACHA20_POLY1305 ] list k
     | Dev : ([ `Tun | `Tap ] * string option) k
     | Dhcp_disable_nbt : flag k
     | Dhcp_dns : Ipaddr.t list k
@@ -409,7 +410,7 @@ module Conf_map = struct
         | `SHA384 -> "SHA384"
         | `SHA512 -> "SHA512")
     in
-    let pp_cipher ppf v = Fmt.string ppf (cipher_to_string v) in
+    let pp_cipher ppf v = Fmt.string ppf (cipher_to_string (v :> cipher)) in
     match (k, v) with
     | Auth, h -> p () "auth %a" pp_digest_algorithm h
     | Auth_nocache, () -> p () "auth-nocache"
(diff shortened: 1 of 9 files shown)


src/packet.ml
```diff
@@ -319,7 +322,16 @@ let encode_tls_data t =
...
   and opt = Cstruct.of_string t.options in
-  Cstruct.concat [ prefix; key_source; opt_len; opt; null; u_p ]
+  let peer_info = match t.peer_info with
+    | None -> Cstruct.empty
+    | Some peer_info ->
+      let len = String.length peer_info in
+      let pi = Cstruct.create (2 + len) in
+      Cstruct.BE.set_uint16 pi 0 len;
+      Cstruct.blit_from_string peer_info 0 pi 0 len;
+      pi
```
hannesm, line 332: we need an additional 0 byte, i.e. 2 byte length, then peer info, then null

hannesm: see #122 for a helper... you also blit peer_info at the beginning, while it should be at offset 2 (since it's [n_0; n_1; data_0; ..; data_n; 0]

src/config.ml
```diff
@@ -140,6 +148,10 @@ module Conf_map = struct
     | Connect_retry : (int * int) k
     | Connect_retry_max : [ `Unlimited | `Times of int ] k
     | Connect_timeout : int k
+    | Data_ciphers
+        : [ `AES_256_CBC | `AES_128_GCM | `AES_256_GCM | `CHACHA20_POLY1305 ]
+          list
+          k
```
hannesm, line 154: it's not clear to me whether we want to allow AES-256-CBC here.

src/config.ml
```diff
@@ -2125,12 +2155,13 @@ let client_generate_connect_options t =
   let excerpt =
     Conf_map.filter
       (function
-        | B (Cipher, _) -> true
+        | B (Cipher, _) -> false (* TODO: only relevant if no NCP is used *)
```
hannesm, line 2158: this TODO should be addressed... AFAICT we have two modes of operation: tls-auth or static key mode. in the former, we do NCP and do not need the cipher in the push_request, in the latter we need it.

question: do we need to support "connecting a client with NCP to a server without NCP?" -- but I struggle to understand how we can distinguish the server not having NCP (and maybe it's not worth it, since NCP is around for > 4 years now)

src/config.ml
```diff
@@ -2125,12 +2155,13 @@ let client_generate_connect_options t =
...
         | B (Comp_lzo, _) -> true
         | B (Tun_mtu, _) -> true
         | B (Link_mtu, _) -> true
         | B (Pull, _) -> true
         | B (Tls_mode, `Client) -> true
+        | B (Auth, _) -> true (* TODO: only relevant if no NCP is used *)
```
hannesm, line 2164: esp, with #36 in mind, I think the auth should be optional. but that can be addressed in a separate PR

src/engine.ml
```diff
@@ -357,7 +357,15 @@ let maybe_kex_client rng config tls =
     let pre_master, random1, random2 = (rng 48, rng 32, rng 32) in
     Config.client_generate_connect_options config >>= fun options ->
     let user_pass = Config.find Auth_user_pass config in
-    let td = { Packet.pre_master; random1; random2; options; user_pass }
+    let peer_info =
+      let ciphers =
+        String.concat ":"
+          (List.map Config.cipher_to_string (Config.get Data_ciphers config))
+      in
+      Some [ "IV_PLAT=mirage"; "IV_CIPHERS=" ^ ciphers; "IV_NCP=2" ]
```
hannesm, line 365: should there be more to it? should it be confgurable whether we do NCP or not (i.e. --disable-ncp)? or is that overthinking?

src/engine.ml
```diff
@@ -376,84 +384,77 @@ let merge_server_config_into_client config tls_data =
...
+        in
+        CHACHA20_POLY1305
+          {
+            my_key = Mirage_crypto.Chacha20.of_secret my_key;
+            my_implicit_iv;
+            their_key = Mirage_crypto.Chacha20.of_secret their_key;
+            their_implicit_iv;
+          }
+  in
+  { my_packet_id = 1l; their_packet_id = 1l; keys }
```
hannesm, line 454: the reason behind the KDF change is that only at the point where the server send the push_reply we know which cipher to use for the data channel. this means we need to delay the keying until we are done with the push_reply.

the nice thing is that now these function are separate and not in the result monad :)

hannesm:
should be tested with static key mode. the server part is as well untested (but let's delay that until later)

src/engine.ml
reynir: Maybe we can send as well `IV_PROTO=0`

reynir: Or maybe it could be interesting with `IV_PROTO=4` signalling "[t]he client expects a push-reply and  the  server may  send this reply without waiting for a push-request first."

hannesm: hmm hmm, but we'll send a push_request!? but eventually this is not needed and would save a roundtrip?

dinosaure: Several options are set if we want to save a roundtrip according to what `openvpn` does:
https://github.com/OpenVPN/openvpn/blob/7c637b35038e4cc7d2f16e07f413cfa2b0f1f97a/src/openvpn/ssl.c#L2054-L2070

dinosaure: So, it's more `IV_PROTO=276` (`REQUEST_PUSH` | `AUTH_PENDING_KW` | `AUTH_FAIL_TEMP`) plus `IV_MTU` than `IV_PROTO=4`

src/miragevpn.mli
```diff
@@ -51,6 +51,10 @@ module Config : sig
         (** [connect-timeout] a.k.a. [server-poll-timeout]
         Timeout a connection attempt to a remote server after [SECONDS].
     *)
+    | Data_ciphers
+        : [ `AES_256_CBC | `AES_128_GCM | `AES_256_GCM | `CHACHA20_POLY1305 ]
+          list
+          k
```
hannesm, line 57: after looking into the openvpn man page and our implementation, I think that "cipher" should only allow the AES_256_CBC (in the configuration file), and data_ciphers should only allow the 3 AEAD ciphers in the configuration file.

the config is also used for the established session, where Cipher need to allow all ciphers (after NCP, the server will tell us which cipher to use, which may well be AEAD).

but also configurations without "cipher" should be valid (if tls-client/tls-server is used).

src/engine.ml
hannesm: thanks for the link, I think for now, the `REQUEST_PUSH` is the only one we can achieve -- since we do not handle temporary authentication failures, neither auth_pending control messages.

the `IV_MTU` we should sent as well (which is a separate key-value from `IV_PROTO`).

src/miragevpn.mli
hannesm: done in 3403569

src/config.ml
hannesm: done in 1bc5456

src/engine.ml
hannesm: I opened #126 to not forget about this

reynir: For compatibility with openvpn 2.4 `IV_NCP=2` signals the client supports AES-{128,256}-GCM and is deprecated in favor of `IV_CIPHERS` according to https://github.com/OpenVPN/openvpn-rfc/blob/bbde60d7f28a5436e21b3f475788f7922e0d9339/openvpn-wire-protocol.xml#L1398-L1399

We can deal with that in another PR as this is already a lot of changes.

src/state.ml
```diff
@@ -66,7 +66,7 @@ type channel_state =
   | Expect_reset
   | TLS_handshake of Tls.Engine.state
   | TLS_established of Tls.Engine.state * key_source
-  | Push_request_sent of Tls.Engine.state * keys
+  | Push_request_sent of Tls.Engine.state * key_source * Packet.tls_data
```
reynir, line 69: I'm slightly confused by this. `Packet.tls_data` contains all the fields of `key_source`, and at first occurrence in engine.ml it seems this data is duplicated?

src/engine.ml
```diff
@@ -602,23 +601,29 @@ let incoming_control_client config rng session channel now op data =
...
+              let cipher = Config.get Cipher config
+              and hmac_algorithm = Config.get Auth config in
+              let keys = kdf session cipher hmac_algorithm key tls_data in
               let channel_st = Established keys in
               Ok (Some ip_config, config, { channel with channel_st }, tls_out)
           | None ->
               (* now we send a PUSH_REQUEST\0 and see what happens *)
               push_request tls' >>| fun (tls'', out) ->
-              let channel_st = Push_request_sent (tls'', keys) in
+              let channel_st = Push_request_sent (tls'', key, tls_data) in
```
reynir, line 612: Here the `tls_data` seems to be the one sent from the server.

src/engine.ml
```diff
@@ -688,6 +693,9 @@ let incoming_control_server is_not_taken config rng session channel _now _ts
   | TLS_established (tls, keys), Packet.Control ->
       let open Result.Infix in
       incoming_tls_without_reply tls data >>= fun (tls', d) ->
+      Option.to_result
+        ~none:(`Msg "TLS established, expected data, received nothing") d
+      >>= fun d ->
```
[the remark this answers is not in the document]
reynir: Ah no `kex_server` used to call `maybe_kdf` and now we need to do the work here.

src/packet.ml
```diff
@@ -385,18 +396,26 @@ let decode_tls_data ?(with_premaster = false) buf =
...
-             data)
-     else ();
-     match (u, p) with "", "" -> None | _ -> Some (u, p))
-  >>| fun user_pass -> { pre_master; random1; random2; options; user_pass }
+     maybe_string "password" buf (p_start + 2) p_len >>= fun p ->
+     let user_pass = match (u, p) with "", "" -> None | _ -> Some (u, p) in
+     let peer_info_start = p_start + 2 + p_len in
+     (if Cstruct.length buf = peer_info_start then Ok None
+      else
+        guard (Cstruct.length buf >= peer_info_start + 2) `Partial >>= fun () ->
```
reynir, line 404: If we want to fully emulate the behavior of openvpn we should return `Ok None` in this case.

reynir: Yes, correct. The question is if this is really sensible, but it is what openvpn does :-)

src/engine.ml
dinosaure: As far as I understand, we delay this `tls_data` to the `Push_request_sent` which will do the `kdf` on it as we do in the `Some _` branch. So what is the issue here?

src/engine.ml
```diff
@@ -602,23 +601,29 @@ let incoming_control_client config rng session channel now op data =
           match Config.(find Ifconfig config) with
```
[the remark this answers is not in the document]
dinosaure: Let's keep it as is and specify the behavior in another PR, we probably should make an issue, this line seems a bit cryptic.

src/engine.ml
reynir: If I remember correctly I was confused with the `tls_data` in `Push_request_sent _` -- I think in some cases it is our `tls_data` while on other cases it's theirs.

dinosaure: I think the trouble come from the `incoming_control_server` and the `incoming_control_client`. But if we keep only the execution path of the `incoming_control_client`, the only way to construct a `Pull_request_sent` is done via what the server give to us. And again, it just matter of a delay to compute `kdf` _after_ `PUSH_REQUEST\0` if it's needed.

src/config.ml
```diff
@@ -1317,16 +1333,32 @@ let a_server =
...
+  | c -> aead_cipher c)
   >>| fun v -> `Entry (B (Cipher, v))
 
+let a_data_ciphers =
+  string "data-ciphers" *> a_whitespace *> a_single_param >>= fun v ->
+  let ciphers = String.split_on_char ':' v |> List.map String.trim in
+  List.fold_left
+    (fun acc c ->
+      acc >>= fun acc ->
+      aead_cipher c >>| fun c -> c :: acc)
```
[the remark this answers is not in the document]
hannesm: thanks. addressed in 1e276ef

src/engine.ml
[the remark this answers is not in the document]
hannesm: is this related to #99? I agree we should open an issue for improve at least the documentation of why this is here.

src/engine.ml
hannesm: I hope e36d0ba398312bc2d074e02843fab7ee0acf2f1c clarifies this, there's own_key_material and peer_key_material now :)

and in `kex_server`, indeed the wrong bits were attached to Push_request_sent.

src/packet.ml
```diff
@@ -385,18 +395,27 @@ let decode_tls_data ?(with_premaster = false) buf =
...
-             (Cstruct.length data) (Cstruct.to_string data) p Cstruct.hexdump_pp
-             data)
-     else ();
-     match (u, p) with "", "" -> None | _ -> Some (u, p))
-  >>| fun user_pass -> { pre_master; random1; random2; options; user_pass }
+     maybe_string "password" buf (p_start + 2) p_len >>= fun p ->
+     let user_pass = match (u, p) with "", "" -> None | _ -> Some (u, p) in
+     let peer_info_start = p_start + 2 + p_len in
+     (* XXX(dinosaure): if we don't have enough to have a peer-info (at least 2 bytes),
+        we just ignore it and return [None]. *)
```
hannesm, line 402: thanks for your comment. I removed the `XXX` -- which I reserve, similar as `TODO` for things that need to be done (and so a `git grep XXX` `git grep TODO` helps to figure out where to look). since this is a very valid comment, let's leave it as is :)

src/state.ml
hannesm: as mentioned earlier, it is "our_key_material" and "peer_key_material", both of which used to derive the actual key. I renamed those, and it should now be clearer.

reynir:
Thanks! I will merge once CI has finished.

PR #122, opened 2023-10-17
refactor how a string is written onto the wire

hannesm:


src/packet.ml
```diff
@@ -300,28 +300,33 @@ let pp_tls_data ppf t =
...
-        let u_l = Cstruct.create 2 and p_l = Cstruct.create 2 in
-        Cstruct.BE.set_uint16 u_l 0 (succ (String.length u));
-        Cstruct.BE.set_uint16 p_l 0 (succ (String.length p));
-        Cstruct.concat
-          [ u_l; Cstruct.of_string u; null; p_l; Cstruct.of_string p; null ]
-  and opt = Cstruct.of_string t.options in
-  Cstruct.concat [ prefix; key_source; opt_len; opt; null; u_p ]
+        Cstruct.append (write_string u) (write_string p))
+      t.user_pass
+  in
```
reynir, line 328: Perhaps we should write two empty wire strings; that's what openvpn does. Presumably so the peer-info can be unambiguously read. See https://github.com/OpenVPN/openvpn/blob/master/src/openvpn/ssl.c#L2221-L2269

src/packet.ml
```diff
@@ -300,28 +300,33 @@ let pp_tls_data ppf t =
...
      in addition to be length-prefixed... *)
-  let null = Cstruct.create 1 in
-  Cstruct.BE.set_uint16 opt_len 0 (succ (String.length t.options));
-  let u_p =
-    match t.user_pass with
-    | None -> Cstruct.empty
-    | Some (u, p) ->
+  let opt = write_string t.options
+  and u_p =
+    Option.fold ~none:Cstruct.empty
```
reynir, line 323:
```suggestion
    Option.fold ~none:(Cstruct.create 4)
```
or 
```suggestion
    Option.fold ~none:(Cstruct.append (wire_string "") (wire_string ""))
```

src/packet.ml
reynir: Also https://github.com/OpenVPN/openvpn/blob/master/src/openvpn/ssl.c#L2377-L2380

reynir (approved):
Thanks! This is nice. I have a somewhat unrelated request that will help with peer-info.

reynir:
Let's save the peer-info related changes for another PR :-)

PR #124, opened 2023-10-17
encode_tls_data: be more nice to allocator

hannesm:


reynir (approved):

src/packet.ml
```diff
@@ -315,17 +315,26 @@ let encode_tls_data t =
...
   in
-  Cstruct.concat [ prefix; key_source; opt; u_p ]
+  (* prefix - 4 zero bytes, key_method
+     pre_master
+     random1
+     random2
+     opt string
+     user string
+     password string
+     NCP
```
reynir, line 334:
```suggestion
     peer-info
```

src/packet.ml
```diff
@@ -315,17 +315,26 @@ let encode_tls_data t =
...
+     let key_source = Cstruct.concat [  ] in
```
reynir, line 335: I don't understand this comment

src/packet.ml
hannesm: indeed. done in c0f72a8

src/packet.ml
hannesm: leftover, removed in c0f72a8

PR #125, opened 2023-10-17
support in tls mode the GCM ciphers, and chacha20; fixes #112

### LINKED ISSUE #112: Support other cipher than AES-256-CBC
there's as well --data-ciphers, and nowadays default is instead AES-256-GCM. Our codebase has AES-256-CBC hardcoded...

hannesm:
in static key mode, only CBC cipher(s) are supported. we use the configuration option "cipher" incorrectly (there should be data-ciphers) -- cipher is only used for static key mode, it seems.

anyways, this works with all 4 ciphers in tls-mode now :D

### DIFF
--- a/src/config.ml
+++ b/src/config.ml
@@ -134,7 +134,8 @@ module Conf_map = struct
           option
           k
     | Ca : X509.Certificate.t list k
-    | Cipher : string k
+    | Cipher
+        : [ `AES_256_CBC | `AES_128_GCM | `AES_256_GCM | `CHACHA20_POLY1305 ] k
     | Comp_lzo : flag k
     | Connect_retry : (int * int) k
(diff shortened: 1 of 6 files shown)


hannesm:
still todo: is compression with tls-mode (and aead ciphers) done?

dinosaure (approved):
Just a note for a futur version of me. `tls-auth` does not do any negotiation about the cipher. However, the client can specify via the `--cipher` argument which cipher it wants to use. On the server side, the cipher specified by the client MUST exist as a `--data-ciphers`.

For instance, if the client choose `CHACHA20-POLY1305` and this cipher is not listed into the `--data-ciphers` on the server-side, the connection will fail (with `AUTH_FAIL`).

Otherwise, this code works with OpenVPN 2.6.6 and all ciphers listed with `tls-auth`. Thanks!

PR #127, opened 2023-10-19
Tls-crypt-v2 (rewrite)

reynir:
This is an alternative to #105. Tls-crypt-v2 requires some restructuring of code. The packet encoding is different enough that it warrants its own separate implementation. Packet decoding is even further removed and is to be done in two stages.

### DIFF
--- a/src/packet.ml
+++ b/src/packet.ml
@@ -253,6 +253,160 @@ let to_be_signed key p =
   | `Control (_, c) -> to_be_signed_control op c
   | `Data _ -> assert false
 
+module Tls_crypt = struct
+  (* Functions to implement:
+     - to_be_signed
+     - encode
+     - decode
+  *)
+  type cleartext_header = {
+    local_session : int64;
+    packet_id : packet_id;
+    timestamp : int32;
+    hmac : Cstruct.t; (* always 32 bytes *)
+  }
+
+  type clear_pkt =
+    [ `Ack of cleartext_header * Cstruct.t | `Control of operation * cleartext_header * Cstruct.t | `Data of Cstruct.t ]
+
+  let hmac_len = Mirage_crypto.Hash.SHA256.digest_size (* 32 *)
+  let hmac_offset = 16
+  let encrypted_offset = hmac_offset + hmac_len
+  let clear_hdr_len = hdr_len hmac_len - 1 (* not including acked msg ids *)
+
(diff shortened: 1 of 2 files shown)


hannesm:
the first two commits are definitively fine, feel free to "dune build @fmt" locally and push them directly to the main branch :)

src/packet.ml
```diff
@@ -44,12 +46,13 @@ let operation_to_int, int_to_operation =
 let pp_operation ppf op =
   Fmt.string ppf
     (match op with
-    | Soft_reset -> "soft reset"
+    | Soft_reset_v2 -> "soft reset"
```
hannesm, line 49:
```suggestion
    | Soft_reset_v2 -> "soft reset v2"
```

src/packet.ml
```diff
@@ -16,22 +16,24 @@ let pp_error ppf = function
...
 
 type operation =
-  | Soft_reset
+  | Soft_reset_v2 (* SOFT_RESET_V2 *)
   | Control
   | Ack
   | Data_v1
-  | Hard_reset_client
-  | Hard_reset_server
+  | Hard_reset_client_v2 (* HARD_RESET_CLIENT_V2 *)
```
hannesm, line 23:
```suggestion
  | Hard_reset_client_v2
```

src/packet.ml
```diff
@@ -16,22 +16,24 @@ let pp_error ppf = function
...
+  | Hard_reset_server_v2 (* HARD_RESET_SERVER_V2 *)
```
hannesm, line 24:
```suggestion
  | Hard_reset_server_v2
```

src/result.ml
```diff
@@ -7,3 +7,8 @@ module Infix = struct
   let ( >>= ) = bind
   let ( >>| ) x f = map f x
 end
+
+module Syntax = struct
+  let ( let* ) = bind
+  let ( let+ ) x f = map f x
+end
```
hannesm, line 14: I like the Syntax (more than the Infix operators) -- maybe we should move the entire codebase to use only the syntax?

reynir: Yea, I would like that as well. Maybe in another PR?

hannesm: yes

hannesm:
thanks a lot

Issue #126, opened 2023-10-19
avoid PUSH_REQUEST roundtrip (send IV_PROTO bit 2), also send IV_MTU


hannesm:
leftovers from #121 #102:

Several options are set if we want to save a roundtrip according to what openvpn does:
https://github.com/OpenVPN/openvpn/blob/7c637b35038e4cc7d2f16e07f413cfa2b0f1f97a/src/openvpn/ssl.c#L2054-L2070

hannesm:
fixed by #129

PR #129, opened 2023-10-19
Delete one roundtrip with `REQUEST_PUSH`

dinosaure:
On top of #121. The last commit save one roundtrip between the client and the server: we don't need anymore to send `PUSH_REQUEST\0` if `pull`/`client` is set.

### DIFF
--- a/src/engine.ml
+++ b/src/engine.ml
@@ -357,6 +357,7 @@ let maybe_kex_client rng config tls =
   if Tls.Engine.can_handle_appdata tls then
     let pre_master, random1, random2 = (rng 48, rng 32, rng 32) in
     Config.client_generate_connect_options config >>= fun options ->
+    let pull = Config.mem Pull config in
     let user_pass = Config.find Auth_user_pass config in
     let peer_info =
       let ciphers =
@@ -365,6 +366,13 @@ let maybe_kex_client rng config tls =
              (Config.get Data_ciphers config))
       in
       Some [ "IV_PLAT=mirage"; "IV_CIPHERS=" ^ ciphers; "IV_NCP=2" ]
+    in
+    let peer_info =
+      match peer_info with
+      | Some peer_info when pull ->
+          Some ("IV_PROTO=4" (* IV_PROTO_REQUEST_PUSH *) :: peer_info)
+      | Some peer_info -> Some ("IV_PROTO=0" :: peer_info)
+      | None -> None
     in
     let td =
       { Packet.pre_master; random1; random2; options; user_pass; peer_info }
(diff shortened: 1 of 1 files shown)


dinosaure:
It solve #126.

src/engine.ml
```diff
@@ -357,6 +357,7 @@ let maybe_kex_client rng config tls =
   if Tls.Engine.can_handle_appdata tls then
     let pre_master, random1, random2 = (rng 48, rng 32, rng 32) in
     Config.client_generate_connect_options config >>= fun options ->
+    let pull = Config.mem Pull config in
```
reynir, line 360: `--client` implies `--tls-client` and `--pull` -- In Config we synthesize this. So this is fine.

src/engine.ml
```diff
@@ -366,6 +367,13 @@ let maybe_kex_client rng config tls =
       in
       Some [ "IV_PLAT=mirage"; "IV_CIPHERS=" ^ ciphers; "IV_NCP=2" ]
     in
+    let peer_info =
+      match peer_info with
+      | Some peer_info when pull ->
+          Some ("IV_PROTO=4" (* IV_PROTO_REQUEST_PUSH *) :: peer_info)
```
reynir, line 373: At some point we probably want to generate this from a bitmap, but for now it's fine.

hannesm (approved):
LGTM - in incoming_control_client, we could `s/key/own_key_material/`

src/engine.ml
```diff
@@ -611,15 +619,24 @@ let incoming_control_client config rng session channel now op data =
...
-              (* now we send a PUSH_REQUEST\0 and see what happens *)
-              push_request tls' >>| fun (tls'', out) ->
-              let channel_st = Push_request_sent (tls'', key, tls_data) in
-              (* first send an ack for the received key data packet (this needs to be
-                 a separate packet from the PUSH_REQUEST for unknown reasons) *)
-              ( None,
-                config,
-                { channel with channel_st },
-                tls_out @ [ (`Ack, Cstruct.empty); (`Control, out) ] )))
+              let pull = Config.mem Pull config in
```
reynir, line 622: Do we otherwise support `--tls-client` without `--pull`?

src/engine.ml
```diff
@@ -611,15 +619,24 @@ let incoming_control_client config rng session channel now op data =
...
+              else
+                (* now we send a PUSH_REQUEST\0 and see what happens *)
+                push_request tls' >>| fun (tls'', out) ->
+                let channel_st = Push_request_sent (tls'', key, tls_data) in
+                (* first send an ack for the received key data packet (this needs to be
+                   a separate packet from the PUSH_REQUEST for unknown reasons) *)
+                ( None,
+                  config,
+                  { channel with channel_st },
+                  tls_out @ [ (`Ack, Cstruct.empty); (`Control, out) ] )))
```
reynir, line 639: Do we actually want to send a push request if we don't `--pull`?

dinosaure: This is, from what I saw, the current behavior of OpenVPN. `IV_PROTO=4` is activated only if `opt->pull` is set to `true`. It permits to keep the "old" behavior for us. Not sure if we want to keep it however.

src/engine.ml
dinosaure: That's the question :+1:. Not sure if it's the case even for OpenVPN.

src/engine.ml
reynir: I haven't looked into PUSH_REQUEST and what it does, but from the name it sounds like something that is only useful with `--pull`. I may be very wrong of course :)

src/engine.ml
hannesm: at least the code follows the old code -- i.e. if we do not send `IV_PROTO=4` then we send a `PUSH_REQUEST` -- and follow the old style handshake.

reynir: I think OpenVPN supports `--tls-client` without `--pull`!

       --pull This option must be used on a client which is  connecting  to  a
              multi-client  server. It indicates to OpenVPN that it should ac‐
              cept options pushed by the server, provided they are part of the
              legal  set  of  pushable options (note that the --pull option is
              implied by --client ).

              In particular, --pull allows the server to push  routes  to  the
              client,  so  you should not use --pull or --client in situations
              where you don't trust  the  server  to  have  control  over  the
              client's routing table.

src/engine.ml
hannesm: indeed, in our state machine (and to behave well with a server), we need to send a push_request and get a push_reply to end up in established. we always do this in the "tls-client" mode (unless ifconfig is specified).

what is the reasoning behind it? to support a point-to-point connection, we don't need any push_request. everything is static (ip configuration, ...).

for a server that can deal with multiple clients, we need the client to have dynamically assigned an ip address (sent by the server, which knows which ips are already used). here a push_request is sensible.

below the line, there may be more configurational choice in terms of whether we need a push_request or not -- but it looks like the whole NCP_CIPHERS needs a push_reply in any case so the server can settle on a cipher to use for the data channel.

src/engine.ml
hannesm: yes, see https://github.com/robur-coop/miragevpn/pull/129#discussion_r1368424228 -- if there's interest (actual deployment, example configuration), we can have such a configuration being tested. for now, I opened #133 to keep track of this

reynir (approved):
I'm a bit fuzzy about what it means to not have `--pull`, but I think this PR at least does not make it 'worse', and looks good otherwise.

PR #130, opened 2023-10-19
app: close fd when read/write failed

hannesm:


dinosaure:
Seems good for me :+1:.

Issue #106, opened 2023-10-09
CI building unikernels should use correct branch


hannesm:
At the moment, each unikernel adds a "pin-depend" on the miragevpn#main repository for the miragevpn package.

This makes the cirrus CI testing the unikernels with the main branch of miragevpn instead of the current branch.

We should change this, maybe we can have a script after `depend` to "rm -rf duniverse/miragevpn && ln -s <??> duniverse/miragevpn` to always use the right branch.

PR #137, opened 2023-10-23
Cirrus fix

hannesm:
experiment for #106


hannesm:
force-pushed after I saw the right failure :) this is ready for being merged, and solves #106

reynir (approved):
Thanks! It would be nice if we could make it work with the mirage tooling, but I remember this setup is not easy with mirage :/

PR #139, opened 2023-10-23
tls-crypt-v2 (again)

reynir:


src/packet.ml
```diff
@@ -430,8 +465,8 @@ module Tls_crypt = struct
           in
```
hannesm, line 465: in the line above, the `encrypted` includes the 2 byte length of wkc -- likely that should be discarded... the split should be `encrypted .. 2 byte wkc len .. wkc`` -- so the middle part (length) can be safely discarded

src/packet.ml
```diff
@@ -386,15 +408,33 @@ module Tls_crypt = struct
...
+    let res = {
+      local_session;
+      packet_id;
+      timestamp;
+      hmac;
+      ack_message_ids;
+      remote_session;
+    }
+    in
+    (res, 1 + arr_len + rs_len)
```
hannesm, line 420: should be `arr_len * packet_id_len` here

src/packet.ml
reynir: The wkc len is part of the authenticated data so we need to preserve it.

hannesm: ok, but should it be part of `encrypted`? (not sure what is done with encrypted afterwards)

reynir:
The client can successfully send `Hard_reset_client_v3` and decrypt and parse the server's `Hard_reset_server_v2`. The client detects a non-monotonic packet id, and this is likely the hmac cookie mechanism described in `doc/tls-crypt-v2.txt`.

src/packet.ml
reynir: No, wkc is not encrypted with the client key (though parts of it is encrypted with the server key). The `wkc_len` is the length of the wkc *including* the `wkc_len` btw.

hannesm: oh, interesting. I wasn't aware the length would include the length field itself. I suspect then there's nothing to do here :)

reynir:
The server response `HARD_RESET_SERVER_V2` has a packet id that signals the server supports hmac cookies, and this has to be worked around it seems. Then the client sends a `CONTROL_V1` and the server reads and prepares a server hello. It then sends an `ACK_V1`, but then for reasons I cannot understand it sends `HARD_RESET_SERVER_V2` and resets the connection. The server log does not seem to give any clues why it resets the connection.

Two (annoying) things I discovered:
- Don't pipe the output of openvpn to `less` (unless using `less +F`): once buffers have been filled it will block the openvpn process and handshakes may time out.
- The `protocol_dump` in openvpn is completely broken for tls-crypt and the output is misleading and wrong.

reynir:
I had passed `None` as message id to `expected_packet` for incoming control packets which meant that
1) The message id was not checked
1) An ACK was never sent by the client
1) The server resending unacked control packets after ~3 seconds

Now the client can successfully connect to a server I set up using tls-crypt-v2! :partying_face:

The PR still needs a bit of a cleanup, and we likely can refactor some common code in tls-crypt-v2 and tls-auth.

src/packet.ml
```diff
@@ -309,7 +303,11 @@ module Tls_crypt = struct
     let op = op_key (operation p) key in
     match p with
     | `Ack hdr -> fst (to_be_signed_header op hdr)
+    | `Control (Hard_reset_client_v3, (hdr, msg_id, _wkc)) ->
+        (* HARD_RESET_CLIENT_V3 is special: the wkc is not considered part of the packet *)
+        to_be_signed_control op (hdr, msg_id, Cstruct.empty)
     | `Control (_, c) -> to_be_signed_control op c
+    | `Data _ -> assert false
```
hannesm, line 310: I think this is fine, but would appreciate a comment why it is fine -- i.e. "data packets are never processed with tls-crypt"

src/packet.ml
```diff
@@ -354,6 +352,7 @@ module Tls_crypt = struct
       match p with
       | `Ack ack -> encode_header ack
       | `Control (op, control) -> encode_control op control
+      | `Data _ -> assert false (* XXX *)
```
hannesm, line 355: again, a comment why this is good would be welcome :)

src/packet.ml
```diff
@@ -364,14 +363,14 @@ module Tls_crypt = struct
     let prefix = encode_protocol proto (Cstruct.lenv [ op_buf; payload ]) in
     let r = Cstruct.concat [ prefix; op_buf; payload ] in
     (* packet, to_encrypt_offset, to_encrypt_length *)
-    (r, Cstruct.length prefix + encrypted_offset, len - encrypted_offset)
+    (r, Cstruct.length prefix + 1 + encrypted_offset, len - encrypted_offset)
```
hannesm, line 366: what about "Cstruct.length op_buf" instead of the 1? this would make it easier to understand.

hannesm: I don't quite understand why it is "encrypted_offset" here -- shouldn't it be related to `payload`?

src/packet.ml
```diff
@@ -395,6 +394,25 @@ module Tls_crypt = struct
         ack_message_ids;
         remote_session;
       }
+    in
+    (res, 1 + (arr_len * packet_id_len) + rs_len)
```
hannesm, line 398: the `1` is the byte of the `arr_len` field?

hannesm:
reviewed packet.ml changes, they look good -- see minor comments above

src/state.ml
```diff
@@ -258,13 +258,28 @@ type tls_auth = {
...
   | Client_tls_auth of { tls_auth : tls_auth; state : client_state }
+  | Client_tls_crypt of {
+      tls_crypt : tls_crypt * Cstruct.t;
+      state : client_state;
+    }
   | Client_static of { keys : keys; state : client_state }
   | Server_tls_auth of { tls_auth : tls_auth; state : server_state }
 
 let pp_state ppf = function
   | Client_tls_auth { state; _ } -> pp_client_state ppf state
```
hannesm, line 280: Should we `Fmt.pf ppf "client tls-auth %a" pp_client_state state` here?

src/engine.ml
```diff
@@ -101,6 +101,21 @@ let hmac_and_out protocol key { hmac_algorithm; my_hmac; _ } (p : Packet.pkt) =
...
+  let iv = Cstruct.sub hmac 0 16 in
+  let ctr = Mirage_crypto.Cipher_block.AES.CTR.ctr_of_cstruct iv in
+  let header = Packet.header p in
+  let p = Packet.with_header { header with Packet.hmac } p in
+  let buf, enc_off, enc_len = Packet.Tls_crypt.encode protocol (key, p) in
+  let encrypted =
+    Mirage_crypto.Cipher_block.AES.CTR.encrypt ~key:my_key ~ctr
+      (Cstruct.sub buf enc_off enc_len)
+  in
+  Cstruct.blit encrypted 0 buf enc_off enc_len;
```
hannesm, line 116: not obvious to me right now why `encrypted` should fit into the bounds returned from `Packet.Tls_crypt.encode` (`enc_off`, `enc_len`). Maybe worth a comment or a guard?

src/engine.ml
```diff
@@ -101,6 +101,21 @@ let hmac_and_out protocol key { hmac_algorithm; my_hmac; _ } (p : Packet.pkt) =
   let p' = Packet.with_header { header with Packet.hmac } p in
   Packet.encode protocol (key, p')
 
+let encrypt_and_out protocol key { my_key; my_hmac; _ } (p : Packet.pkt) =
+  let to_be_signed = Packet.Tls_crypt.to_be_signed key p in
+  let hmac = Mirage_crypto.Hash.SHA256.hmac ~key:my_hmac to_be_signed in
+  let iv = Cstruct.sub hmac 0 16 in
```
hannesm, line 107: is the 16 the block size of AES that is used? if yes, should we use `Mirage_crypto.AES.CTR.block_size` instead?

hannesm:
engine.ml also looks fine (though I skipped over some parts, a diff view isn't very readable).

I suggest to test (tls-auth & tls-crypt) [I guess this has been done already], then merge -- we can always improve / refactor later :)

src/packet.ml
reynir: There are two occurences of `encrypted_offset`:

1) The `payload` includes the cleartext part of the header, the encrypted part of the header (acks, remote session), the encrypted data (if any) and the cleartext wkc (if `HARD_RESET_CLIENT_V3`). `encrypted_offset` is the offset into `payload` that is (to be) encrypted.

2) In the second occurence we subtract from the length of the payload the length of the cleartext part of the payload (`encrypted_offset`).

The wkc is not considered part of the payload as such even though in the code we do. It's a choice between encoding without wkc, encrypting and then fixing the proto length, or pass the wkc as part of the payload and then returning the range to encrypt. We may need to reconsider this if we want to implement `P_CONTROL_WKC_V1` as in that case we will have a non-empty encrypted data as well as cleartext wkc.

src/packet.ml
reynir: Yes

src/packet.ml
reynir: I added comments to explain why. It would be nice to not have this case, but I failed to make it type check. Probably it's a case of adding `(_ :> Packet.pkt)` in the right places.

src/engine.ml
reynir: If I remember correctly in the documentation they specify it as the first half of the hmac, or the 128 most significant bits (without specifying endianness).

src/engine.ml
reynir: Hmm, do you mean in `(Cstruct.sub buf enc_of enc_len)`? Or `encrypted` having length `enc_len`? The former is by construction in `Packet.Tls_crypt.encode`. The latter is due to AES in CTR-mode not adding any padding or overhead. I can add a comment for both points, but for the former I don't know if I should put it in the function or at the call site(s).

reynir:
I did test tls-crypt-v2 and tls-auth and both seems to work. I didn't test very extensively.

src/engine.ml
hannesm: fine then

src/packet.ml
hannesm: that may deserve a comment or not, your take :)

src/packet.ml
hannesm: thanks, and your comment on the `encrypted_offset` binding makes a lot of sense and helps me to understand this better.

src/engine.ml
reynir: I think it is not the easiest code to follow, but I find it hard to make it more clearer. Comments would help!

PR #143, opened 2023-10-23
remove Result.Infix module, replace all call sites with Result.Syntax

hannesm:
as proposed in #139


reynir:
I rebased on main, replaced `Result.map` with `let+`, and found and removed an identity function application.

reynir:
Force pushed as I had missed a parenthesis, and the identity function was not an identity function after all.

PR #145, opened 2023-10-24
adjust log level of ping (incoming and outgoing) messages to [info]

hannesm:


hannesm:
I decided to use `debug` level in both cases instead.

PR #147, opened 2023-10-26
Actually parse tls-version-max

reynir:


hannesm:
thanks for spotting this

PR #155, opened 2023-11-01
Refactor init_channel, rekey for tls-crypt-v2

reynir:
- Refactor init_channel and remove init_channel_tls_crypt
- Use Cstruct.empty as hmac placeholder
- Pass Packet.t over a separate keyid and Packet.pkt
- Implement tls-crypt-v2 rekeying
- Store Packet.t in out_packets rather than serialized Cstruct.t. This can allow for retransmission with new replay packet id.

### DIFF
--- a/src/engine.ml
+++ b/src/engine.ml
@@ -93,14 +93,14 @@ let compute_hmac key p hmac_algorithm hmac_key =
   Mirage_crypto.Hash.mac hmac_algorithm ~key:hmac_key tbs
 
 let hmac_and_out protocol { hmac_algorithm; my_hmac; _ }
-    ((key, p) : [< Packet.pkt_ack | Packet.pkt_control ] Packet.t) =
+    (key, (p : [< Packet.ack | Packet.control ])) =
   let hmac = compute_hmac key p hmac_algorithm my_hmac in
   let header = Packet.header p in
   let p' = Packet.with_header { header with Packet.hmac } p in
   Packet.encode protocol (key, p')
 
 let encrypt_and_out protocol { my_key; my_hmac; _ }
-    ((key, p) : [< Packet.pkt_ack | Packet.pkt_control ] Packet.t) =
+    (key, (p : [< Packet.ack | Packet.control ])) =
   let to_be_signed = Packet.Tls_crypt.to_be_signed key p in
   let hmac = Mirage_crypto.Hash.SHA256.hmac ~key:my_hmac to_be_signed in
   let iv = Cstruct.sub hmac 0 16 in
@@ -900,7 +900,7 @@ type error =
   | `Mismatch_my_session_id of transport * Packet.header
   | `Msg_id_required_in_fresh_key of transport * int * Packet.header
   | `Different_message_id_expected_fresh_key of transport * int * Packet.header
-  | `Bad_mac of t * Cstruct.t * Packet.pkt Packet.t
+  | `Bad_mac of t * Cstruct.t * Packet.t
   | `No_transition of channel * Packet.operation * Cstruct.t
   | `Tls of
     [ `Alert of Tls.Packet.alert_type | `Eof | `Fail of Tls.Engine.failure ]
--- a/src/engine.ml
+++ b/src/engine.ml
@@ -92,13 +92,15 @@ let compute_hmac key p hmac_algorithm hmac_key =
   let tbs = Packet.to_be_signed key p in
   Mirage_crypto.Hash.mac hmac_algorithm ~key:hmac_key tbs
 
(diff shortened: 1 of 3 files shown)


src/engine.ml
```diff
@@ -95,13 +92,13 @@ let compute_hmac key p hmac_algorithm hmac_key =
...
 
-let hmac_and_out protocol key { hmac_algorithm; my_hmac; _ } (p : Packet.pkt) =
+let hmac_and_out protocol { hmac_algorithm; my_hmac; _ } ((key, p) : Packet.t) =
   let hmac = compute_hmac key p hmac_algorithm my_hmac in
   let header = Packet.header p in
   let p' = Packet.with_header { header with Packet.hmac } p in
   Packet.encode protocol (key, p')
 
-let encrypt_and_out protocol key { my_key; my_hmac; _ } (p : Packet.pkt) =
+let encrypt_and_out protocol { my_key; my_hmac; _ } ((key, p) : Packet.t) =
```
reynir, line 101: We need to ensure we don't call this with ``(_, `Data _)`` as `Packet.Tls_crypt.encode` will assert false.

src/engine.ml
```diff
@@ -1105,26 +1086,32 @@ let maybe_init_rekey s =
...
+  | Client_tls_crypt { state = Ready; tls_crypt } ->
+      let session, channel, out = init_channel () in
+      let state = Client_tls_crypt { state = Rekeying channel; tls_crypt } in
+      let out = encrypt_and_out s.session.protocol (fst tls_crypt) out in
       ({ s with state; session }, [ out ])
   | Client_static _ ->
       (* there's no rekey mechanism in static mode *)
       (s, [])
-  | _ ->
+  | Client_tls_auth _ | Client_tls_crypt _ | Server_tls_auth _ ->
```
reynir, line 1114: This change would have caught the missing tls-crypt-v2 rekeying

reynir:
I managed to refine the types and remove a few `assert false` branches. The types now enforce we only hmac and encrypt control packets (``[ `Control of _ | `Ack of _ ]``), and only control packets that are not ACKs are added to `out_packets`, our retransmit map. I'm very pleased with this change.

src/packet.ml
```diff
@@ -430,25 +427,20 @@ module Tls_crypt = struct
     ({ local_session; packet_id; timestamp; hmac }, clear_hdr_len)
 end
 
-type pkt =
-  [ `Ack of header | `Control of operation * control | `Data of Cstruct.t ]
+type pkt_ack = [ `Ack of header ]
+type pkt_control = [ `Control of operation * control ]
+type pkt = [ pkt_ack | pkt_control | `Data of Cstruct.t ]
+type 'a t = int * ([< pkt ] as 'a)
```
hannesm, line 433: what about naming it `type ack` and `type control`?

hannesm: and I have not checked it, but the `type pkt` and `type 'a t` -- can we get rid of one of them? AFAICT `Packet.t` is only refered to once, in `` Bad_mac of t * Cstruct.t * Packet.t`... maybe worth to remove it, rename `type pkt` above to `type t`, and have a `int * Packet.t` where needed?

src/engine.ml
```diff
@@ -70,16 +70,13 @@ let header session hmac_algorithm transport timestamp =
...
-    Cstruct.create hmac_len
-  in
   let my_packet_id = Int32.succ packet_id in
   ( { session with my_packet_id },
     { transport with last_acked_message_id },
     {
       Packet.local_session = session.my_session_id;
-      hmac;
+      hmac = Cstruct.empty;
+      (* placeholder *)
```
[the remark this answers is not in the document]
hannesm: just remove it ;) another option would be to make that `hmac` optional (and provide a function that constructs such a thing, where hmac is an optional labeled argument).

in the end, the hmac being allowed to be absent is in line with #36

hannesm:
this looks very nice, I only put some minor nits.

src/packet.ml
reynir: `type control` already existed, but it was only used once so I inlined it as well as `pp_control`. I pushed a commit addressing your comments. I kept `type t` but without the type variable.

src/engine.ml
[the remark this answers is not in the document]
reynir: I don't think it makes sense to make `hmac` optional - how we designed it we need to create a packet before we can compute the hmac -- so we will never know up front what the hmac is I'm afraid. An upside is we don't allocate a cstruct that we never intend to use now.

hannesm:
good to merge

PR #156, opened 2023-11-06
Rename packet_id to replay_id; message_id to packet_id

hannesm:


hannesm:
since it's a renaming, I'd be keen to merge it rather sooner than later... @reynir what are your thoughts about the renaming? does "replay ID" and "packet ID" make it clearer (also in respect to https://git.robur.coop/robur/openvpn-spec/issues/6)

hannesm:
we could as well rename the new `packet id` (former message id) as `sequence number`? WDYT?

reynir:
I think I would like to avoid 'packet id'. I like sequence number. In the openvpn code the replay packet id (replay id) code mostly lives in `packet_id.[ch]` while the packet id (sequence number) lives mostly in reliable.c **but uses** the `packet_id_type` type from `packet_id.h`. It is highly confusing when reading the code that packet_id is used for both so I would prefer if we can avoid that term.

I like replay_id even if it's a half-truth as the timestamp is considered part of the replay packet id.

reynir:
I force pushed the suggested `sequence number` change. I also found some more occurences of  "packet[_ ]id" and renamed them to `replay_id` or `replay packet id`.

reynir (approved):

src/packet.ml
```diff
@@ -269,29 +269,29 @@ module Tls_crypt = struct
   let clear_hdr_len = hdr_len hmac_len - 1 (* not including acked msg ids *)
 
   let to_be_signed_header ?(more = 0) op header =
-    (* op_key ++ session_id ++ packet_id ++ timestamp ++ ack_len ++ acks ++ remote_session ++ msg_id *)
-    let acks_len = List.length header.ack_message_ids * packet_id_len
+    (* op_key ++ session_id ++ sequence_number ++ timestamp ++ ack_len ++ acks ++ remote_session ++ msg_id *)
```
reynir, line 272: Sorry, I made a mistake here
```suggestion
    (* op_key ++ session_id ++ replay_id ++ timestamp ++ ack_len ++ acks ++ remote_session ++ sequence_number *)
```

src/engine.ml
```diff
@@ -861,40 +865,40 @@ let expected_packet session transport data =
...
+      (Int32.equal session.their_replay_id hdr.Packet.replay_id)
+      (`Non_monotonic_replay_id (transport, hdr))
   in
   let+ () =
     opt_guard
-      (Int32.equal transport.their_message_id)
+      (Int32.equal transport.their_sequence_number)
       msg_id
-      (`Non_monotonic_message_id (transport, msg_id, hdr))
+      (`Non_monotonic_sequence_number (transport, msg_id, hdr))
```
reynir, line 875: We should rename msg_id with `sequence_number`, and rename `Packet.message_id` as well.

PR #158, opened 2023-11-07
Hmac cookie support

### LINKED ISSUE #151: tls-crypt-v2 hmac cookie support
In #139 support for tls-crypt-v2 is added. To introduce fewer moving parts at a time hmac cookie support was not implemented. It is desirable to implement hmac cookie support as servers can be configured not to accept such clients as it's a denial of service vector. It shouldn't be too difficult to implement.

reynir:
This adds support for hmac cookies in tls-crypt-v2 for the client. The client signals support and parses flags (encoded using TLV) in the server's response. The client then sends instead of a `Control` packet a `Control_wkc` packet that is essentially the same just with another op code and with `wKc` appended in cleartext.

What I found interesting is that the server responds with (replay) packet id of `0x0f000001` if *we* don't use that (replay) packet id. Presumably to signal support for hmac cookies even if the client didn't ask. This comment says a *peer* can use that packet id value suggesting the server is allowed to use this way as well:
https://github.com/OpenVPN/openvpn/blob/f53f06316dbb804128fc5cbee1d8edb274ce81df/src/openvpn/ssl_pkt.h#L302-L308

I have not tested if this is actually the case.

Another observation I had was that the OpenVPN implementation does *not* send `CONTROL_WKC_V1` when using TCP. Presumably this is because a handshake has already been performed in the TCP layer, and hmac cookies are less important for TCP. This implementation does *not* do anything differently for TCP or UDP.

I need to test this works with TCP as well.

I also fix UDP which I broke in an earlier PR.

Fixes #151.

### DIFF
--- a/src/engine.ml
+++ b/src/engine.ml
@@ -575,7 +575,9 @@ let incoming_control_client config rng session channel now op data =
   match (channel.channel_st, op) with
   | Expect_reset, (Packet.Hard_reset_server_v2 | Packet.Soft_reset_v2) ->
       (* for rekey we receive a soft_reset -- a bit alien that we don't send soft_reset *)
-      (* we reply with ACK + TLS client hello! *)
+      (* we reply with embedded ACK + TLS client hello! *)
+      (* NOTE: For tls-crypt-v2 hmac cookies it is important we don't send a
+         dedicated ACK as we need to ensure the Control_wkc arrives first *)
       let tls, ch =
         let authenticator =
           match Config.find Ca config with
@@ -1852,6 +1854,7 @@ let incoming ?(is_not_taken = fun _ip -> false) state buf =
                                 act )
                           | _ -> assert false))))
         in
+        (* Invariant: [linger] is always empty for UDP *)
         if Cstruct.is_empty linger then Ok (state, out, act)
         else multi linger (state, out, act)
   in
--- a/src/engine.ml
+++ b/src/engine.ml
@@ -1364,7 +1364,7 @@ let wrap_hmac_control now ts mtu session tls_auth key transport outs =
   in
   (session, transport, List.rev outs)
 
-let wrap_tls_crypt_control now ts mtu session (tls_crypt, wkc) hmac_cookie key
+let wrap_tls_crypt_control now ts mtu session (tls_crypt, wkc) needs_wkc key
     transport outs =
   let now_ts = ptime_to_ts_exn now in
   let acks = bytes_of_acks transport in
@@ -1372,7 +1372,7 @@ let wrap_tls_crypt_control now ts mtu session (tls_crypt, wkc) hmac_cookie key
      packet, the Control_wkc has room for the /cleartext/ wkc, and fix the
      packet length afterwards *)
   let session, transport, maybe_out, outs =
-    match (hmac_cookie, outs) with
+    match (needs_wkc, outs) with
     | true, (`Control, data) :: rest ->
         let l = min (mtu - Cstruct.length wkc) (Cstruct.length data) in
         let data, data' = Cstruct.split data l in
@@ -1746,7 +1746,7 @@ let incoming ?(is_not_taken = fun _ip -> false) state buf =
                     Packet.Tls_crypt.decode_decrypted_control cleartext
                       decrypted
                   in
-                  let* hmac_cookie_supported =
+                  let* needs_wkc =
                     match op with
                     | Hard_reset_server_v2 ->
                         Packet.decode_early_negotiation_tlvs data
@@ -1804,8 +1804,8 @@ let incoming ?(is_not_taken = fun _ip -> false) state buf =
                       in
                       let session, transport, encs =
                         wrap_tls_crypt_control (state.now ()) (state.ts ())
-                          my_mtu state.session (tls_crypt, wkc)
-                          hmac_cookie_supported key ch.transport out'
+                          my_mtu state.session (tls_crypt, wkc) needs_wkc key
+                          ch.transport out'
                       in
                       let out = out @ encs
                       and ch = { ch with transport }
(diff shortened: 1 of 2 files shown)


src/engine.ml
```diff
@@ -1332,9 +1332,40 @@ let wrap_hmac_control now ts mtu session tls_auth key transport outs =
...
+let wrap_tls_crypt_control now ts mtu session (tls_crypt, wkc) needs_wkc key
+    transport outs =
   let now_ts = ptime_to_ts_exn now in
-  let outs = split_control mtu outs in
+  (* If we reply with hmac cookie we must split such that the first control
+     packet, the Control_wkc has room for the /cleartext/ wkc, and fix the
+     packet length afterwards *)
+  let session, transport, maybe_out, outs =
+    match (needs_wkc, outs) with
+    | true, (`Control, data) :: rest ->
```
reynir, line 1343: This is brittle. We need to ensure somehow that we don't send an `ACK` for the server's `HARD_RESET_SERVER_V2` as that the server will not know how to decrypt the ACK.

src/engine.ml
```diff
@@ -1332,9 +1332,40 @@ let wrap_hmac_control now ts mtu session tls_auth key transport outs =
...
+        (* First we encrypt *)
+        let out = encrypt_and_out session.protocol tls_crypt (key, p) in
+        (* Then we append wkc and fix the length if TCP *)
+        let len =
+          Cstruct.length out
+          - match session.protocol with `Tcp -> 2 | `Udp -> 0
+        in
+        let out = Cstruct.append out wkc in
+        let proto = Packet.encode_protocol session.protocol len in
+        Cstruct.blit proto 0 out 0 (Cstruct.length proto);
```
reynir, line 1361: I'm not too fond of this hack. We can't do the same hack as for `HARD_RESET_CLIENT_V3` where we pretend the wKc is part of the packet data and special case the encoder. On the other hand maybe we should implement the same hack for `HARD_RESET_CLIENT_V3` instead of using two different hacks.

src/engine.ml
```diff
@@ -1332,9 +1332,40 @@ let wrap_hmac_control now ts mtu session tls_auth key transport outs =
...
+        (session, transport, Some out, rest)
+    | true, _ ->
+        Log.warn (fun m ->
+            m "wrap_tls_crypt_control: expected control to append wkc");
+        (session, transport, None, outs)
```
reynir, line 1366: Maybe `Log.err` and `assert false`?  See the above comment about brittleness.

src/engine.ml
```diff
@@ -1681,10 +1722,7 @@ let incoming ?(is_not_taken = fun _ip -> false) state buf =
                     | Expect_reset, Hard_reset_server_v2 ->
                         let hdr = Packet.header p in
                         Log.warn (fun m ->
-                            m "Hard_reset_client_v2 data: %a" Cstruct.hexdump_pp
-                              data);
-                        Log.warn (fun m ->
-                            m "fixing their_packet_id: %lx" hdr.packet_id);
+                            m "fixing their_packet_id: %08lx" hdr.packet_id);
                         { state.session with their_packet_id = hdr.packet_id }
```
reynir, line 1726: It turns out we *don't* fix `their_packet_id` if we send the special packet id ourselves. I'm not sure what to do about this code...

src/engine.ml
```diff
@@ -1770,7 +1809,8 @@ let incoming ?(is_not_taken = fun _ip -> false) state buf =
                                 act )
                           | _ -> assert false))))
         in
-        multi linger (state, out, act)
+        if Cstruct.is_empty linger then Ok (state, out, act)
+        else multi linger (state, out, act)
```
reynir, line 1813: Invariant: `linger` is always empty for UDP.

src/engine.ml
reynir: I documented in the relevant places (hopefully) that we should not send out dedicated ACKs for this reason.

src/engine.ml
reynir: Switched to `Log.err (...); assert false`

src/engine.ml
reynir: After some thinking I think it would be interesting to always send a `CONTROL_WKC_V1` with an empty body and then pretend the `wKc` is the body and employ the same hack we use for `HARD_RESET_CLIENT_V3`.

reynir:
I struggle a bit rebasing this on main due to 45f5ac95329b83fcf32a48e8e93328ff6de27f4f.

hannesm:
Should we look at it together? Which function is troublesome for the rebase?

reynir:
I decided on a way to merge it. What was uncovered was that the PR did not properly split control packets: It would be split once for `CONTROL_WKC_V1` and the remainder in a `CONTROL_V1` which works as long as the wkc + control data fits in `2*mtu`.  
The second struggle was whether to put ACKs in `CONTROL_WKC_V1`. I decided not to, but I am now questioning that. If the control data and wkc fits in one mtu then we don't ack I think.

I think it is technically legal to send a `CONTROL_WKC_V1` packet with no control data and only the wKc. I might try do that instead. Then the ACKs can be put into the first `CONTROL_V1` packet after the `CONTROL_WKC_V1`

src/engine.ml
```diff
@@ -1364,9 +1366,42 @@ let wrap_hmac_control now ts mtu session tls_auth key transport outs =
...
+     packet, the Control_wkc has room for the /cleartext/ wkc, and fix the
+     packet length afterwards *)
+  let session, transport, maybe_out, outs =
+    match (needs_wkc, outs) with
+    | true, (`Control, data) :: rest ->
+        let l = min (mtu - Cstruct.length wkc) (Cstruct.length data) in
+        let data, data' = Cstruct.split data l in
+        let rest =
+          if Cstruct.is_empty data' then rest else (`Control, data') :: rest
+        in
```
reynir, line 1383:
```suggestion
        (* To ensure we send back ACKs we unconditionally add another control *)
        let rest = (`Control, data') :: rest in
```
I need to check that an empty control packet doesn't cause issues, but maybe this is one way to fix the issue mentioned.

PR #162, opened 2023-11-08
remove Packet.cipher_block_size; also re-implement State.mtu

hannesm:


src/state.ml
```diff
@@ -313,44 +313,46 @@ let pp ppf t =
...
-  let left = data mod Packet.cipher_block_size in
-  let r =
-    if left = pred Packet.cipher_block_size then data - not_yet_padded_payload
-    else data - succ left - not_yet_padded_payload
-  in
-  assert (r > 0);
-  r
+  and compress = session.compress in
+  match Config.get Cipher config with
+  | `AES_256_CBC ->
```
reynir, line 324: Hmm I wonder how this interacts with tls-crypt which uses a fixed cipher and hmac algorithm :/

src/state.ml
```diff
@@ -313,44 +313,46 @@ let pp ppf t =
...
+      let static_key_mode = Config.mem Secret config in
+      let ts = if static_key_mode then 4 (* timestamp *) else 0 in
+      let block_size = Mirage_crypto.Cipher_block.AES.CBC.block_size in
+      let hmac = Config.get Auth config |> Mirage_crypto.Hash.digest_size in
```
reynir, line 328: Same here for tls-crypt

reynir (commented):
I have some comments regarding tls-crypt. It seems to me the mtu computation is for the data channel - I wonder if we need separate `mtu_data` and `mtu_control`?

In any case I find the new code more readable, and it addresses the configurable cipher.

src/state.ml
hannesm: I was wondering as well -- and guess we need both a data_mtu and a control_mtu -- since the former is used for any payload to be transferred via the wire; and the second needs to take the hmac etc. into account -- even if AEAD is used, the control channel (even for tls-auth) is using the auth. This means that the match on cipher is wrong.

src/state.ml
```diff
@@ -354,6 +354,41 @@ let mtu config session =
       in
       tun_mtu - hdr
 
+let control_mtu config state session =
+  (* we assume to have a tun interface and the server send us a tun-mtu *)
+  let tun_mtu =
+    match Config.find Tun_mtu config with
+    | None -> 1500 (* TODO "client_merge_server_config" should do this! *)
+    | Some x -> x
+  and compress = session.compress in
```
reynir, line 363: Is compression also applied for the control channel?

src/state.ml
```diff
@@ -354,6 +354,41 @@ let mtu config session =
...
+  match state with
+  | Client_static _ -> assert false
+  | Client_tls_auth _ | Server_tls_auth _ ->
+      (* here, the hash is used from auth *)
+      let hmac = Config.get Auth config |> Mirage_crypto.Hash.digest_size in
+      let not_yet_padded_payload = Packet.id_len + if compress then 1 else 0 in
+      let hdrs =
+        hmac + 1 (* key /op *) + if session.protocol = `Tcp then 2 else 0
```
reynir, line 371: Maybe ``1 (* key / op *) + Packet.hdr_len hmac + if session.protocol = `Tcp then 2 else 0``

src/state.ml
```diff
@@ -354,6 +354,41 @@ let mtu config session =
...
+      let pad =
+        let res = data mod block_size in
+        if res = pred block_size then 0 else succ res
+      in
+      let r = data - pad - not_yet_padded_payload in
+      assert (r > 0);
+      r
+  | Client_tls_crypt _ ->
+      (* there's the cleartext header *)
+      let hdr = Packet.Tls_crypt.clear_hdr_len in
```
reynir, line 386: Maybe we should use instead `Packet.hdr_len`. Neither takes into account non-empty ack list and remote session.

reynir (commented):
I struggle a bit with how we can compute the exact control packet header length when there can be variable number of acks. Maybe one way forward is to add an upper bound of number of ACKs - openvpn uses 4 IIRC due to a bug/limitation in another implementation, and then be conservative in `control_mtu`.

hannesm:
Hold on, the MTU computation is used for `split_control` --> we would like to know how much actual space we have for a control payload. The actual space is computed by taking the tun_mtu and subtracting the cipher/hmac header.

So, on the wire, how does a control packet look?
- TCP 2 bytes length, for UDP no length
- 1 byte key/op

now, we distinguish tls_auth:
- hmac
- header (as you mention, variable length)
- sequence_number
- payload

tls_crypt (IV is taken from the hmac)
- hmac
- header (as you say, variable length)
- sequence number
- payload

et voila, we don't need separate cases. and indeed, I read that the max number of ACKs should be 8 - we can also use 4. and fundamentally we don't need any block size :)

src/state.ml
hannesm: it is not, fixed in 3938ec1f3f4e5c5822bbdca243970f84ab764b75

src/state.ml
hannesm: yes, that's morally what we have in 3938ec1f3f4e5c5822bbdca243970f84ab764b75

src/state.ml
hannesm: true, this is done in 3938ec1f3f4e5c5822bbdca243970f84ab764b75 with max_acks being 8.

hannesm:
For the max_acks, I wonder what we really want:
- option (a) do not account in control_mtu, but rather in the split_control?
- option (b) have a limit and enforce this (though a different implementation may decide on a different limit)

reynir:
I'm very happy with this PR now. I'm inclined to merging this and create an issue on further refining mtu calculation.

reynir (approved):

hannesm:
I thought let's take (a), and be done with it :)

src/engine.ml
```diff
@@ -1274,30 +1276,53 @@ let check_control_integrity err key p hmac_algorithm hmac_key =
...
-              let l = min mtu (Cstruct.length data) in
-              let data, data' = Cstruct.split data l in
-              datas (data :: acc) data'
-          in
-          List.map (fun data -> (`Control, data)) (datas [] data)
-      | ((`Ack | `Reset_server | `Reset), _) as p -> [ p ])
-    outs
+let split_control ~acks mtu outs =
+  (* only the first control/ack packet is carrying acks *)
+  List.rev
```
reynir, line 1281: I think you forgot to remove this one in the latest commit :-)

hannesm: Hmm? we're folding left, and accumulate "new_thing :: acc", so in the end we want to reverse to retain the order.

I removed in [62c3a61](https://github.com/robur-coop/miragevpn/pull/162/commits/62c3a6198e1dfd72b046210b0fffccea20bdd521) the inner List.rev when we actually split some control data.

reynir: Ah my bad

PR #164, opened 2023-11-21
add bsd 2 clause license

hannesm:
The reasoning behind it is: we're all fine with that license, and we enjoy writing that code so much that it's fine for anyone to use, extend, do whatever they want with it. While we're also fine with AGPL, this may prevent people from using this software (namely in companies whose legal departments are allergic to GPL/AGPL [due to unresolved issues of these licenses]).

Please @cfcs could you approve (or disapprove) this license change, since you contributed quite a lot back than.

Thanks a lot.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,23 @@
+Copyright (c) 2019 robur.coop
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without modification,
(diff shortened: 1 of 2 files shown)


reynir (approved):

reynir:
Great! @dinosaure also approved on IRC; merging.

PR #169, opened 2023-11-27
implement tls-crypt

### LINKED ISSUE #166: Implement tls-crypt
With tls-crypt-v2 implemented it is relatively straight forward to implement plain tls-crypt. While tls-crypt-v2 has more desirable properties especially in larger deployments it would be great to implement tls-crypt as that is seeing use in deployments.

hannesm:
fixes #166 (still needs testing (the tls-crypt-v2 part seems to work fine)


hannesm:
how can this happen? does tls-crypt again have some direction and we're failing on that? in your configuration, do you use reverse directions for client and server?

hannesm:
Maybe time to refactor this PR (esp. `Engine.incoming`)?

src/engine.ml
```diff
@@ -1596,89 +1637,64 @@ let find_channel state key op =
...
-                      and ch = { ch with transport } in
-                      Ok (set_ch state ch, out, act))
-              | ( _control_op,
+                  let acts =
+                    Option.fold payload ~none:acts ~some:(fun p ->
+                        `Payload p :: acts)
+                  in
+                  (state, out, acts)
+              | Packet.Hard_reset_client_v3, _ ->
+                  Error (`No_transition (ch, op, payload))
```
hannesm, line 1690: this is slightly dubious, but preserves the earlier semantics

src/engine.ml
```diff
@@ -1815,26 +1779,70 @@ let incoming ?(is_not_taken = fun _ip -> false) state buf =
...
                       and ch = { ch with transport } in
-                      Ok (set_ch state ch, out, act))
-              | Packet.Hard_reset_client_v3, Client_tls_crypt _ ->
-                  Error (`No_transition (ch, op, payload))
-              | _control_op, Client_tls_crypt { tls_crypt = tls_crypt, wkc; _ }
-                -> (
-                  let* cleartext, off =
+                      match p with
+                      | `Ack _ -> Ok (set_ch state ch, out, acts)
+                      | `Control (_, (_, _, data)) -> (
```
hannesm, line 1797: would be nice to sync the tls_auth and tls_crypt* cases:
- verify_incoming data
- process packet
- sign/encrypt outgoing

this way, the logic (such as ``match out' with [] -> [ (`Ack, Cstruct.empty) ] | xs -> xs``) wouldn't need to be duplicated

bonus points if we merge the tls_crypt_v2 with the `wkc` into the same shape

src/state.ml
```diff
@@ -409,6 +411,84 @@ let channel_of_keyid keyid s =
...
+          t with
+          state = Server_tls_auth { state = Server_ready; tls_auth };
+          lame_duck;
+        },
+        None )
+  | Client_tls_auth _ -> assert false
+  | Client_tls_crypt _ -> assert false
+  | Client_tls_crypt_v2 _ -> assert false
+  | Server_tls_auth _ -> assert false
+  | Client_static _ -> assert false
```
hannesm, line 490: I understand these have been there before, but I'm nevertheless uneasy with these assertions.

src/engine.ml
hannesm: esp. that there haven't been any verification of the packet taken place makes it dubious (opens an attacker to send us a hard_reset_client_v3 to result in an error state).

reynir: Ah yes for UDP we should maybe just drop it

reynir:
Good news is both our tlsauth and tls-crypt-v2 configs seem to still work after the major refactoring.

src/engine.ml
reynir: Right. This is maybe okay for TCP, but for UDP we should at least try to authenticate before returning ``Error `No_transition _``.

Maybe we should do a TCP/UDP error semantics overhaul at some point...

src/engine.ml
```diff
@@ -1690,167 +1697,74 @@ let incoming ?(is_not_taken = fun _ip -> false) state buf =
...
+                      op decrypted
                   in
                   let* needs_wkc =
-                    match op with
-                    | Hard_reset_server_v2 ->
+                    match p with
+                    | `Control (Packet.Hard_reset_server_v2, (_, _, data)) ->
                         Packet.decode_early_negotiation_tlvs data
                     | _ -> Ok false
                   in
```
reynir, line 1767: A downside is we try to decode the early negotiation TLVs even for tls-crypt. We could match on `wkc_op, op` and only do the decoding if the wkc is present (tls-crypt-v2).

src/engine.ml
```diff
@@ -1460,15 +1500,15 @@ let wrap_hmac_control now ts mtu session tls_auth key transport outs =
...
+let wrap_tls_crypt_control now ts mtu session tls_crypt wkc_opt needs_wkc key
     transport outs =
   let now_ts = ptime_to_ts_exn now in
   (* If we reply with hmac cookie we must split such that the first control
      packet, the Control_wkc has room for the /cleartext/ wkc, and fix the
      packet length afterwards *)
   let session, transport, maybe_out, outs =
-    match (needs_wkc, outs) with
-    | true, (`Control, data) :: rest ->
+    match (wkc_opt, needs_wkc, outs) with
```
reynir, line 1510: We could unify `wkc_opt` and `needs_wkc` so we pass always `None` to `wrap_tls_crypt_control` when (current) `needs_wkc` is false

reynir:
I observed AEAD decrypt failures on both the OpenVPN™ server and the client when using tls 1.3 on my own debian server with both tls-crypt *and* tls-crypt-v2. Using tls 1.2 works fine however. I tested with the main branch and I can reproduce the same behavior (except for tls-crypt of course).

Interestingly I can't reproduce against the OpenVPN™ server running on FreeBSD.

hannesm:
since the ekm stuff is unrelated to tls-crypt, let's: not consider that as a stopper from merging this PR; and disable ekm in engine.ml for now (until we find the time & energy to debug it in more depth).

src/engine.ml
hannesm: I think the UDP error semantics needs revision, but this is fine for now.

hannesm:
I'm reasonably happy with this PR now. We should at some point think about the UDP error handling (where we should ignore bad & short packets etc., those that have invalid mac, ...).

PR #170, opened 2023-11-27
Pkcs12

hannesm:
this works to a degree... there's something special now, since pkcs12 files are usually DER-encoded, but when inlined have a base64 encoding (as specified in the openvpn man page)

### DIFF
--- a/src/engine.ml
+++ b/src/engine.ml
@@ -165,7 +165,7 @@ let tls_crypt config =
           wkc,
           force_cookie )
 
-let client config ts now rng =
+let client ?pkcs12_password config ts now rng =
   let open Result.Syntax in
   let current_ts = ts () in
   let config =
@@ -182,6 +182,52 @@ let client config ts now rng =
         done;
         let remotes = Array.to_list remotes in
         Config.add Remote remotes config
+  in
+  let* config =
+    match Config.find Pkcs12 config with
+    | None -> Ok config
+    | Some p12 -> (
+        let* pass =
+          Option.to_result ~none:(`Msg "missing pkcs12 password")
+            pkcs12_password
+        in
+        let* stuff = X509.PKCS12.verify pass p12 in
+        let cert =
+          List.filter_map
+            (function `Certificate c -> Some c | _ -> None)
+            stuff
+        and key =
+          List.filter
+            (function
+              | `Private_key _ | `Decrypted_private_key _ -> true | _ -> false)
+            stuff
+        in
+        match (cert, key) with
+        | certs, ([ `Private_key pk ] | [ `Decrypted_private_key pk ]) -> (
+            let key_fp =
+              X509.Public_key.fingerprint (X509.Private_key.public pk)
+            in
+            match
+              List.filter
+                (fun cert ->
+                  let cert_pubkey = X509.Certificate.public_key cert in
+                  Cstruct.equal (X509.Public_key.fingerprint cert_pubkey) key_fp)
+                certs
+            with
+            | [ cert ] ->
+                Ok
+                  Config.(
+                    add Tls_key pk (add Tls_cert cert (remove Pkcs12 config)))
+            | _ ->
+                Error
+                  (`Msg "PKCS12: couldn't find matching cert for private key"))
+        | _, keys ->
+            Error
+              (`Msg
+                (Fmt.str
+                   "expected exactly one private key in PKCS12, found %u \
+                    private keys"
+                   (List.length keys))))
   in
   let* action, state =
     match Config.get Remote config with
--- a/src/engine.ml
+++ b/src/engine.ml
@@ -202,27 +202,32 @@ let client ?pkcs12_password config ts now rng =
               | `Private_key _ | `Decrypted_private_key _ -> true | _ -> false)
             stuff
         in
-        (* TODO: allow for certs as --ca or --extra-certs *)
         match (cert, key) with
-        | [ cert ], ([ `Private_key pk ] | [ `Decrypted_private_key pk ]) -> (
+        | certs, ([ `Private_key pk ] | [ `Decrypted_private_key pk ]) -> (
             let key_fp =
               X509.Public_key.fingerprint (X509.Private_key.public pk)
             in
-            let cert_pubkey = X509.Certificate.public_key cert in
-            if Cstruct.equal (X509.Public_key.fingerprint cert_pubkey) key_fp then
-              Ok
-                Config.(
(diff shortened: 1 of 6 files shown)


hannesm:
what I get from our out-of-band access:
```
2023-11-27T17:46:46+01:00: [WARNING] [ovpn.engine] ignoring bad packet non monotonic sequence number 4 in local 13634293449916875355 replay_id 9 timestamp 1701105653 hmac 
                                                   cb ca 4c ec 04 91 84 a0  32 f5 0e ab 0e 2a e9 ff
                                                   9f 36 1d 5f 4c 3a a5 97  b3 51 c6 57 0a aa fe 3e
                                                    ack  remote  
                                                   (state my packet 5
                                                   their packet 5 (acked 5)
                                                   out 1)
```

hannesm:
(this is after the TLS handshake being completed, so we actually go a pretty long way :D)

src/engine.ml
```diff
@@ -183,6 +183,52 @@ let client config ts now rng =
...
+                (fun cert ->
+                  let cert_pubkey = X509.Certificate.public_key cert in
+                  Cstruct.equal (X509.Public_key.fingerprint cert_pubkey) key_fp)
+                certs
+            with
+            | [ cert ] ->
+                Ok
+                  Config.(
+                    add Tls_key pk (add Tls_cert cert (remove Pkcs12 config)))
+            | _ ->
```
reynir, line 221: Apparently `--pkcs12` can also be used instead of `--ca`. So I wonder if `--pkcs12` and `--ca` are mutually exclusive. I will investigate.
```suggestion
            match
              List.partition
                (fun cert ->
                  let cert_pubkey = X509.Certificate.public_key cert in
                  Cstruct.equal (X509.Public_key.fingerprint cert_pubkey) key_fp)
                certs
            with
            | [ cert ], _ :: _ as cas ->
                Ok
                  Config.(
                    add Ca cas (add Tls_key pk (add Tls_cert cert (remove Pkcs12 config))))
            | _ ->
```

reynir (approved):
This looks good. I need to look up how `--ca` and `--pkcs12` interact or if they are mutually exclusive.

src/engine.ml
hannesm: hmm, but there may be as well a certificate chain in the PKCS12 (to be send to the server)... I'm not entirely sure how OpenVPN handles this tbh.

also, the CA for the server may be different than the CA for the user authentication (i.e. a server may have a public dns name, and use let's encrypt certs -- while the client gets a certificate from a private CA).

reynir: It seems `--pkcs12` and `--ca` are not mutually exclusive https://github.com/OpenVPN/openvpn/blob/a1cb1b47b138b9f654cd0bca5de6d08dbca61888/src/openvpn/options.c#L2952-L2978 even if `--pkcs12` and `--capath` are mutually exclusive (odd!). I am not sure what the semantics are.

reynir: Yes, this is very confusing. There is as well `--extra-certs` which are certificates in the chain of the local certificate and not used for authenticating the remote (while `--ca` can be used for both). It would make more sense if `--pkcs12` can be used instead of `--extra-certs` (and not `--ca`).

hannesm:
Thanks for your investigation - I'm happy with the changes you made. Could you autoformat (`dune build @fmt --auto`) and then it's ready to be merged.

hannesm:
but unfortunately, in our out-of-band PKCS12 we have 2 certificates and one private key -- so your [50a0fc3](https://github.com/robur-coop/miragevpn/pull/170/commits/50a0fc3136c8729a70d4713d067f8961656dd09f) results in early erroring out (I now remember that this was the reason why I did the List.filter with the check which cert is the one to use).

reynir:
Yes, I will revert and document that we don't add certs from pkcs12 to --ca

hannesm:
Maybe we should as well warn about these certificates that we ignore (by printing the subject (and issuer?)).

src/engine.ml
```diff
@@ -183,6 +183,61 @@ let client config ts now rng =
...
+                  Cstruct.equal (X509.Public_key.fingerprint cert_pubkey) key_fp)
+                certs
+            with
+            | [ cert ], extra_certs ->
+                (* OpenVPN adds the certs to [Ca] if it is not present;
+                   we don't as it is easily surprising behavior *)
+                if extra_certs <> [] then
+                  Log.warn (fun m ->
+                      m "Extra certificates found in PKCS12 file%s: %a"
+                        (if Config.mem Ca config then ""
```
reynir, line 223: I'm not sure our config validator allows for `--pkcs12` without `--ca` hmm.

reynir (approved):

src/engine.ml
```diff
@@ -183,6 +183,64 @@ let client config ts now rng =
...
+          Option.to_result ~none:(`Msg "missing pkcs12 password")
+            pkcs12_password
+        in
+        let* stuff = X509.PKCS12.verify pass p12 in
+        let certs =
+          List.filter_map
+            (function `Certificate c -> Some c | _ -> None)
+            stuff
+        and keys =
+          List.filter
```
hannesm, line 200:
```suggestion
          List.filter_map
```

src/engine.ml
```diff
@@ -183,6 +183,64 @@ let client config ts now rng =
...
+            (function
+              | `Private_key _ | `Decrypted_private_key _ -> true | _ -> false)
```
hannesm, line 202:
```suggestion
              | `Private_key pk | `Decrypted_private_key pk -> Some pk | _ -> None)
```

src/engine.ml
```diff
@@ -183,6 +183,64 @@ let client config ts now rng =
...
+            stuff
+        in
+        match keys with
+        | [ (`Private_key pk | `Decrypted_private_key pk) ] -> (
```
hannesm, line 206:
```suggestion
        | [ pk ] -> (
```

src/engine.ml
hannesm: it does, ca is completely optional:
```OCaml
          match Config.find Ca config with
          | None ->
              Log.warn (fun m ->
                  m "not authenticating certificate (missing CA)");
              fun ?ip:_ ~host:_ _ -> Ok None
          | Some ca ->
              Log.info (fun m ->
                  m "authenticating with CA %a"
                    Fmt.(list ~sep:(any "\n") X509.Certificate.pp)
                    ca);
```

hannesm:
now, this is fine (also with my example pkcs12) -- great to merge (I'd squash-merge since there's no point to keep the individual commits)

PR #171, opened 2023-11-28
remove context from error messages to make them clearer

hannesm:


src/engine.ml
```diff
@@ -69,6 +69,10 @@ let header session transport timestamp =
   let ack_sequence_numbers =
     acked_sequence_numbers transport.last_acked_sequence_number
   in
+  Log.debug (fun m -> m "last ack %lu, seq %lu ack: %a"
+                transport.last_acked_sequence_number
+                transport.their_sequence_number
+                Fmt.(list ~sep:(any ", ") int32) ack_sequence_numbers);
```
reynir, line 75: Ideally we print the ack list as uint32 (which fmt doesn't have sadly)
```suggestion
                Fmt.(list ~sep:(any ", ") (fun ppf -> pf ppf "%lu")) ack_sequence_numbers);
```

src/engine.ml
```diff
@@ -881,14 +885,15 @@ let expected_packet session transport data =
      both from their side, and acks from our side *)
   let* () =
     guard
-      (Int32.equal session.their_replay_id hdr.Packet.replay_id)
-      (`Non_monotonic_replay_id (transport, hdr))
+      (Int32.unsigned_compare session.their_replay_id hdr.Packet.replay_id <= 0)
+      (`Non_monotonic_replay_id (session.their_replay_id, hdr.Packet.replay_id))
   in
+  Log.debug (fun m -> m "received %a" Fmt.(option ~none:(any "no") int32) sn);
```
reynir, line 891:
```suggestion
  Log.debug (fun m -> m "received %a" Fmt.(option ~none:(any "no") (fun ppf -> pf ppf "%lu")) sn);
```

reynir: Or maybe this slightly shorter version (though less readable IMO)
```suggestion
  Log.debug (fun m -> m "received %a" Fmt.(option ~none:(any "no") (Fun.flip pf "%lu")) sn);
```

src/engine.ml
```diff
@@ -912,8 +919,8 @@ let expected_packet session transport data =
 type error =
   [ Packet.error
   | Lzo.error
-  | `Non_monotonic_replay_id of transport * Packet.header
-  | `Non_monotonic_sequence_number of transport * int32 option * Packet.header
+  | `Non_monotonic_replay_id of int32 * int32
+  | `Non_monotonic_sequence_number of int32 * int32 option
```
reynir, line 923: This could be without an option because this error only makes sense when the sequence number is present. We need to use something else than `opt_guard` then.
```suggestion
  | `Non_monotonic_sequence_number of int32 * int32
```

reynir (approved):
Thank you! I think this is very nice and cuts out irrelevant details from error messages.

src/engine.ml
hannesm: indeed true, but this requires some further changes to construct the value then.

hannesm:
I fixed the code and autoformatted, fine to squash-merge.

reynir:
Thank you! You were faster than me :-) I will merge in a second.

src/engine.ml
```diff
@@ -881,14 +886,19 @@ let expected_packet session transport data =
      both from their side, and acks from our side *)
   let* () =
     guard
-      (Int32.equal session.their_replay_id hdr.Packet.replay_id)
-      (`Non_monotonic_replay_id (transport, hdr))
+      (Int32.unsigned_compare session.their_replay_id hdr.Packet.replay_id <= 0)
```
hannesm, line 889: this actually changes behaviour, but imho this is correct -- there may be lost packets, and all we care is that the replay_id increases (so it should be greater or equal the one we expect)

Issue #35, opened 2020-03-28
Support key pinning


cfcs:
```
       --verify-hash hash [algo]
              Specify SHA1  or  SHA256  fingerprint  for  level-1  cert.   The
              level-1  cert  is  the  CA (or intermediate cert) that signs the
              leaf certificate, and is one removed from the  leaf  certificate
              in  the direction of the root.  When accepting a connection from
              a peer, the level-1 cert fingerprint must match hash or certifi‐
              cate verification will fail.  Hash is specified as XX:XX:... For
              example:

                  AD:B0:95:D8:09:C8:36:45:12:A9:89:C8:90:09:CB:13:72:A6:AD:16

              The algo flag can be either SHA1 or SHA256.  If not provided, it
              defaults to SHA1.
```

See also: https://github.com/OpenVPN/openvpn/pull/105

hannesm:
I stepped into this issue, but unfortunately our X.509 API is slightly different -- where our fingerprint authenticators want to verify the hash of the leaf certificate, here in OpenVPN the specification is the fingerprint of the level-1 certificate (i.e. the certificate signing the leaf).

reynir:
`--verify-hash` is deprecated since at least 2.6, and 2.6 has `--peer-fingerprint`:

     --peer-fingerprint args
               Specify a SHA256 fingerprint or list of  SHA256  fingerprints
               to  verify the peer certificate against. The peer certificate
               must match one of the fingerprint or certificate verification
               will fail. The option can also be inlined

            Valid syntax:

               peer-fingerprint AD:B0:95:D8:09:...

            or inline:

               <peer-fingerprint>
               00:11:22:33:44:55:66:77:88:99:aa:bb:cc:dd:ee:ff:00:11:22:33:44:55:66:77:88:99:aa:bb:cc:dd:ee:ff
               11:22:33:44:55:66:77:88:99:aa:bb:cc:dd:ee:ff:00:11:22:33:44:55:66:77:88:99:aa:bb:cc:dd:ee:ff:00
               </peer-fingerprint>

            When the --peer-fingerprint option is used, specifying a CA with
            --ca or --capath is optional. This allows the he  --peer-finger‐
            print  to  be used as alternative to a PKI with self-signed cer‐
            tificates for small setups. See the examples section for such  a
            setup.

reynir:
I can't quite figure out how `--peer-fingerprint` and `--ca` interact. Neither from the man page nor the code.

[...]

PR #174, opened 2023-11-28
Peer fingerprint

reynir:
This needs some work in order to rebase it. I forgot about this branch.

Implements #35.


hannesm:
in config.ml parse_next we need another `"peer-fingerprint" -> Ok `Peer_fingerprint`. Maybe worth to revise (only have the string -> polyvar once) and add test cases for this new configuration directive!?

reynir:
Ugh yes this is the second time (out of two) that I/we fall into this trap. I will look into it tomorrow. I pushed a quick fix.

PR #176, opened 2023-11-29
Fix <connection> test

reynir:
And also test for comments inside a <connection> block.


reynir:
I pushed a fix for `<pkcs12>` as well. It may a bit too permissive with whitespace. I'm not sure...

PR #184, opened 2023-12-06
split the initial client action from the big variant actions

hannesm:


reynir (approved):
Thanks!

Update: will merge once the CI is finished.

PR #185, opened 2023-12-06
Add test application

reynir:
This application acts as a VPN client that doesn't set up a tun device. Instead, it sends ICMPv4 echo requests to the gateway and reads ICMPv4 echo reply and ignores all other packets. This is useful for testing without touching local routes and configured ip addresses.

This implementation also takes a different approach than miragevpn-client-lwt. It also serves (for me) as an exercise in using our API. Notably, it doesn't use `Lwt.async`. It however blocks on transmit (sending packets to the VPN server) - this may be a poor choice. It also implements a more explicit state machine with three states: connecting, connected and established.
- connecting is self-explanatory.
- connected is when we are connected (have a file descriptor) but have not yet established the VPN connection (no ifconfig)
- established is when we are connected and the VPN connection is established (have both open file descriptor and ifconfig).
This is in contrast to miragevpn-lwt-client where the file descriptor is stored in a mutable `file_descr option` field.

One interesting discovery is that a `` `Connect _ `` action can occur in the established state - it is not preceded by a `` `Disconnect `` action. The semantics of the actions and events could use a brief description in miragevpn.mli.

### DIFF
--- a/app/dune
+++ b/app/dune
@@ -23,6 +23,32 @@
   cstruct-lwt))
 
 (executable
+ (name miragevpn_client_notun)
+ (public_name miragevpn-client-notun)
+ (package miragevpn)
+ (modules miragevpn_client_notun)
+ (libraries
+  miragevpn
+  lwt
+  lwt.unix
+  cmdliner
+  fmt.tty
+  logs
+  logs.fmt
+  logs.cli
+  fmt.cli
+  ipaddr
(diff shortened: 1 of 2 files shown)


hannesm:
CI is happy, I'm fine with merging this sooner than later. And if we spot bugs, we'll fix them :)

PR #187, opened 2023-12-06
Reset linger on connection failure

### LINKED ISSUE #186: Reset linger on connection failed
When we get a `` `Connection_failed `` event we should reset the `linger` field so we don't carry over garbage to our next connection attempt.
https://github.com/robur-coop/miragevpn/blob/a5dca73bcb5c41fa8c357abd4f6d4832c22be72c/src/engine.ml#L1924-L1927

reynir:
This ensures we don't keep around partial packets from previous TCP connections. Fixes #186.

### DIFF
--- a/src/engine.ml
+++ b/src/engine.ml
@@ -1869,7 +1869,7 @@ let retransmit timeout ts transport =
   in
   ({ transport with out_packets }, out)
 
-let resolve_connect_client client t ts s ev =
+let resolve_connect_client make_state t ts s ev =
   let open Result.Syntax in
   let remote, next_remote =
     let remotes = Config.get Remote t.config in
@@ -1901,13 +1901,13 @@ let resolve_connect_client client t ts s ev =
       in
       (* We reset [linger] to enforce the invariant that [linger] is empty when
          we are connecting (or resolving) *)
-      Ok ({ t with linger = Cstruct.empty; state = client state }, action)
+      Ok ({ t with linger = Cstruct.empty; state = make_state state }, action)
   in
   match (s, ev) with
   | Resolving (idx, _, retry), `Resolved ip ->
       (* TODO enforce ipv4/ipv6 *)
       let endp = match remote idx with _, port, dp -> (ip, port, dp) in
-      let t = { t with state = client (Connecting (idx, ts, retry)) } in
+      let t = { t with state = make_state (Connecting (idx, ts, retry)) } in
       Ok (t, Some (`Connect endp))
   | Resolving (idx, _, retry), `Resolve_failed ->
       let+ t, action = next_or_fail t idx retry in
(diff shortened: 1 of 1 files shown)


src/engine.ml
```diff
@@ -1883,7 +1883,7 @@ let resolve_connect_client config ts s ev =
     | `Times m -> m > r
     | `Unlimited -> false
   in
-  let next_or_fail idx retry =
+  let next_or_fail t idx retry =
```
reynir, line 1886: Here we shadow `t`. The reason is because we return `next_or_fail`, and when it is called the state may have changed, at least judging from the structure of the code.

src/engine.ml
```diff
@@ -1892,24 +1892,27 @@ let resolve_connect_client config ts s ev =
...
+            (Connecting (idx', ts, retry'), `Connect (ip, port, dp))
+      in
+      Ok ({ t with linger = Cstruct.empty; state = client state }, action)
   in
   match (s, ev) with
   | Resolving (idx, _, retry), `Resolved ip ->
       (* TODO enforce ipv4/ipv6 *)
       let endp = match remote idx with _, port, dp -> (ip, port, dp) in
-      Ok (Connecting (idx, ts, retry), Some (`Connect endp))
+      let t = { t with state = client (Connecting (idx, ts, retry)) } in
```
hannesm, line 1908: here we don't need to clean up the linger? I understand that the invariant we try to establish is that when we're in "connecting" state, there is no linger, is this correct?

reynir: Yes, we want an invariant that when the state is `Connecting` (or `Resolving`) linger is empty.

hannesm:
Overall, this looks good. I'm still hesitant in respect to passing partially applied functions around, since that makes the code harder to understand. On the other side, it removes code duplication.

We could (in a separate PR) push the "client_state" / "server_state" out of the record where it is combined with the keys... That may make the code easier to understand...

Feel free to merge, minor comments above.

src/engine.ml
hannesm: let's write that down somewhere then :)

src/engine.ml
```diff
@@ -1869,7 +1869,7 @@ let retransmit timeout ts transport =
   in
   ({ transport with out_packets }, out)
 
-let resolve_connect_client config ts s ev =
+let resolve_connect_client client t config ts s ev =
```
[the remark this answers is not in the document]
reynir: I chose `make_state` as it makes a `state` out of a `client_state`.

src/engine.ml
reynir: Yes. It's not immediately obvious how to refactor `next_or_fail`. I arguably made things worse by passing `client` (now `make_client`) outside of the function(s) it's defined in.

PR #189, opened 2023-12-07
Merge tls crypt2

reynir:
I had some opinions on #188 and did my changes here. CC @dinosaure


dinosaure (approved):

src/config.ml
```diff
@@ -1042,22 +1017,22 @@ let inline_payload element =
...
 let a_tls_crypt_v2_client_payload force_cookie =
-  ( inline_pem_payload tls_crypt_v2_client_pem_name "tls-crypt-v2" >>= fun cs ->
-    if Cstruct.length cs >= 256 + Mirage_crypto.Hash.SHA256.digest_size then
-      return cs
-    else fail "bad tls-crypt-v2 client key size" )
-  >>| fun cs ->
-  let key = Cstruct.(sub cs 0 64, sub cs 64 64, sub cs 128 64, sub cs 192 64) in
-  let wkc = Cstruct.shift cs 256 in
-  B (Tls_crypt_v2_client, (key, wkc, force_cookie))
+  let line = take_till (( = ) '\n') <* char '\n' in
```
reynir, line 1020: I missed that before, but maybe we should use `a_newline`:
```suggestion
  let line = take_till (function '\n' | '\r' -> true | _ -> false) <* a_newline in
```
The reason is to support `\r\n` newlines.

src/miragevpn.mli
```diff
@@ -177,11 +177,11 @@ module Config : sig
     | Tls_version_max : Tls.Core.tls_version k
     | Tls_cipher : Tls.Ciphersuite.ciphersuite list k
     | Tls_ciphersuite : Tls.Ciphersuite.ciphersuite13 list k
-    | Tls_crypt : (Cstruct.t * Cstruct.t * Cstruct.t * Cstruct.t) k
+    | Tls_crypt : Tls_crypt.Tls_crypt.t k
```
reynir, line 180: I would like to get rid of `Tls_crypt.Tls_crypt` -- maybe `Tls_crypt.t`?

src/config.ml
```diff
@@ -1098,8 +1073,11 @@ let a_secret =
 
 let a_tls_crypt_payload str =
   let open Result.Syntax in
+  let lines = String.split_on_char '\n' str in
```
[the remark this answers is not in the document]
reynir: I implemented this, but it didn't make a difference in the end due to a quirk in `hex_of_lines` -- will make another comment on that.

src/tls_crypt.ml
```diff
@@ -49,6 +55,32 @@ let pem_of_lines ~name seq =
...
+  | header :: hex_and_footer when header = "-----BEGIN " ^ name ^ "-----" -> (
+      let hex_and_footer = List.trim hex_and_footer in
+      let footer = List.hd (List.rev hex_and_footer) in
+      if footer <> "-----END " ^ name ^ "-----" then
+        error_msgf "Invalid %s" name
+      else
+        let hex = List.chop hex_and_footer in
+        let buf = Buffer.create 0x100 in
+        let f str =
+          for i = 0 to (String.length str / 2) - 1 do
```
reynir, line 73: If the number of characters in a line is odd the last character will be ignored. Lucky for us this means carriage returns will **usually** be ignored. However, if a line ends with a space **and** a carriage return this will cause trouble.

The OpenVPN implementation seems to ignore all whitespace. However, it seems they ignore the last hex digit if there is an odd number of total hex digits(!)
https://github.com/OpenVPN/openvpn/blob/f015643fe23d7847ad45b7763f31bfc6baed2159/src/openvpn/crypto.c#L1257-L1289

PR #191, opened 2023-12-11
refactor: separate state from control_crypto

hannesm:
this simplifies some pattern matches, and makes the code slightly clearer


reynir (approved):
This is very nice!

reynir:
CI error is unrelated.

PR #192, opened 2023-12-11
Unify tls-auth and tls-crypt codepaths more

reynir:
We now have instead of hmac_and_out and encrypt_and_out a function wrap_and_out that decides by inspecting control_crypto how to encrypt and authenticate.

Similarly, we now have wrap_control for both tls-auth and tls-crypt. This allows us to merge handle_client_{tls_auth,tls_crypt} into handle_client.

I don't like that it introduces `assert false` in `Static _` cases.

We should also be able to merge the non-decoding parts of `incoming` in a similar way. This may require refactoring code into function(s).

### DIFF
--- a/src/engine.ml
+++ b/src/engine.ml
@@ -127,6 +127,12 @@ let encrypt_and_out protocol { my; _ }
   Cstruct.blit encrypted 0 buf enc_off enc_len;
   buf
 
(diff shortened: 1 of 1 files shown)


hannesm:
to me this looks great

PR #193, opened 2023-12-11
refactor: simplify incoming

hannesm:
unify the tls_auth and tls_control cases by introducing the validate_control and wrap_control functions


hannesm:
rebased on top of #192

hannesm:
I fixed the code, should now be better... I dislike the `wrap_control` and `wrap_tls_crypt_control` and redefinition of `wrap_control` -- we should find better names)

hannesm:
previous comment is addressed in 488d499

reynir (approved):

Issue #183, opened 2023-12-04
UDP failure behaviour of incoming packets


hannesm:
in tls-auth/tls-crypt/tls-crypt-v2 mode, we should allow some bad UDP packets to not lead to error. Especially stuff where the hmac is bad or which are badly encoded packets should just be ignored, since that may just be a random computer injecting that packet. (this is in `Engine.incoming`)

also, for static mode, we can check what the failure behaviour is (should also be slightly different between udp and tcp).

hannesm:
this has been fixed by #202

PR #194, opened 2023-12-11
use a polyvar

hannesm:
further refactoring, making `incoming` also stricter -- this will need to change for UDP mode (but we need to change stuff for UDP anyways, see #183)

### DIFF
--- a/src/engine.ml
+++ b/src/engine.ml
@@ -989,6 +989,7 @@ type error =
   | `Mismatch_my_session_id of int64 * int64
   | `Bad_mac of t * Cstruct.t * Packet.t
   | `No_transition of channel * Packet.operation * Cstruct.t
+  | `No_channel of int
   | `Tls of
     [ `Alert of Tls.Packet.alert_type | `Eof | `Fail of Tls.Engine.failure ]
   | `Msg of string ]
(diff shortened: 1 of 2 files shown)


src/state.ml
```diff
@@ -393,9 +393,9 @@ let channel_of_keyid keyid s =
             Some (channel, set)
         | _ -> None)
 
-let transition_to_established t =
-  match (t.control_crypto, t.state) with
-  | (Tls_auth _ | Tls_crypt _), Client (Handshaking _) ->
+let transition_to_established t control_crypto =
+  match (control_crypto, t.state) with
+  | #control_tls, Client (Handshaking _) ->
```
reynir, line 398: Thinking out loud: we don't ever really deconstruct or use `control_crypto`. But we pass it on because we can sort of ensure it's a tls client (or server) this way.

src/engine.ml
```diff
@@ -260,107 +259,55 @@ let client ?pkcs12_password config ts now rng =
...
   in
-  match
-    (tls_auth config, tls_crypt config, tls_crypt_v2 config, secret config)
-  with
-  | Error e, Error _, Error _, Error _ -> Error e
-  | Error _, Error _, Error _, Ok (my_key, my_hmac, their_key, their_hmac) ->
-      (* in static key mode, only CBC is allowed *)
-      assert (Config.get Cipher config = `AES_256_CBC);
-      let keys =
+  let* control_crypto =
```
reynir, line 262: Yes! It is nice that this is factored out and we don't have to repeat the other initialization code!

reynir (approved):

src/state.ml
hannesm: turns out, we can as well just drop it. it's purpose right now is to specify at the type level that `transition_to_established` will only be executed with Tls_auth / Tls_crypt

reynir: Yes, we can drop it. It got me thinking if we should have again a `Client_static` that only allows the client states that are relevant. `Handshaking` and `Rekeying` are not valid states for a static client. On the other hand this might reintroduce some pain points. I will do some experiments.

PR #196, opened 2023-12-12
mirage: remove global reference conn_est, use a est_switch in type conn instead

hannesm:


hannesm:
we merged this manually

PR #197, opened 2023-12-12
Engine.handle: return payload explicitely as a list of Cstruct.t, action as option

hannesm:


hannesm:
CI is green, we should nevertheless do a review (suggesting to use https://github.com/robur-coop/miragevpn/pull/197/files?w=1 to ignore whitespace changes)

This partially reverts a97a924295472efe7307b8fd7d2e23aee0c9672f (since handle returns an action option again, and payload is a cstruct.t list).

Also, note that the disconnect event is unused and can be removed subsequently

src/engine.ml
```diff
@@ -1634,31 +1634,44 @@ let incoming ?(is_not_taken = fun _ip -> false) state control_crypto buf =
...
+  let+ s', out, payloads, act_opt =
+    multi (Cstruct.append state.linger buf) (state, [], [], None)
   in
-  let acts' = List.rev acts in
   Log.debug (fun m -> m "out state is %a" State.pp s');
   Log.debug (fun m ->
-      m "%d outgoing packets (%d bytes)" (List.length out) (Cstruct.lenv out));
-  Log.debug (fun m -> m "actions %a" Fmt.(list ~sep:(any "@ ") pp_action) acts');
-  (s', out, acts')
+      m "%u outgoing packets (%d bytes)" (List.length out) (Cstruct.lenv out));
```
reynir, line 1669:
```suggestion
      m "%u outgoing packets (%u bytes)" (List.length out) (Cstruct.lenv out));
```

src/engine.ml
```diff
@@ -1634,31 +1634,44 @@ let incoming ?(is_not_taken = fun _ip -> false) state control_crypto buf =
...
+  Log.debug (fun m ->
+      m "%u payloads (%d bytes)" (List.length payloads) (Cstruct.lenv payloads));
```
reynir, line 1671:
```suggestion
      m "%u payloads (%u bytes)" (List.length payloads) (Cstruct.lenv payloads));
```

reynir (approved):

reynir:
I added a warning when we "drop" an action in `incoming`. The warning could be more specific as the action is ever only `` `Established (ip_config, mtu) ``, but matching on that gets annoying due to the variant no longer being covariant (which can be solved by coercion).

I also tested with the notun application and it still works.

hannesm:
Thanks, I also manually checked with a97a924295472efe7307b8fd7d2e23aee0c9672f and this looks fine. :)

PR #200, opened 2023-12-18
Ignore reneg* = 0

reynir:
We could as well remove the config values if they are zero. But it may be easier to use zero as a special value as we need to negotiate these values with the server (or client).

The special meaning of 0 is only properly documented in the man page for `--reneg-bytes` but the code shows that is applies to `--reneg-pkts` and `--reneg-sec` too:
https://github.com/OpenVPN/openvpn/blob/8656b85c7324fc9ae7f10a9f37227a58766aae33/src/openvpn/ssl.c#L2884-L2893


reynir:
CI failure is unrelated (again a 32 bit capnproto "fix" affecting unrelated packages!)

PR #201, opened 2023-12-18
Config.client_generate_connect_options: only emit "auth" if tls-auth is used

hannesm:
Previously, "auth SHA1" was sent to servers when tls-crypt{,v2} was used

### DIFF
--- a/src/config.ml
+++ b/src/config.ml
@@ -2278,7 +2278,11 @@ let client_generate_connect_options t =
         | B (Link_mtu, _) -> true
         | B (Pull, _) -> true
         | B (Tls_mode, `Client) -> true
-        | B (Auth, _) -> true
+        | B (Auth, _) ->
+            not
+              (Conf_map.mem Tls_crypt t
+              || Conf_map.mem Tls_crypt_v2_client t
+              || Conf_map.mem Tls_crypt_v2_server t)
         | _ -> false)
       t
   in
--- a/src/config.ml
+++ b/src/config.ml
@@ -2278,7 +2278,8 @@ let client_generate_connect_options t =
         | B (Link_mtu, _) -> true
         | B (Pull, _) -> true
         | B (Tls_mode, `Client) -> true
-        | B (Auth, _) -> Conf_map.mem Tls_auth t
+        | B (Auth, _) ->
+            Conf_map.mem Tls_auth t
             (* Assumption: we only support AEAD ciphers in tls clients. Thus
                [Auth] option is ignored for the data channel, and for
                tls-crypt(-v2) the hmac is hardcoded. For tls-auth we still want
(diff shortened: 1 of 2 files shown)


src/config.ml
```diff
@@ -2278,7 +2278,11 @@ let client_generate_connect_options t =
         | B (Link_mtu, _) -> true
         | B (Pull, _) -> true
         | B (Tls_mode, `Client) -> true
-        | B (Auth, _) -> true
+        | B (Auth, _) ->
```
reynir, line 2281: We should send `auth` also if we use non-AEAD cipher. If the assumption is we always use AEAD ciphers we should document this assumption IMO.

hannesm: we should document that invariant, indeed. From my perspective, the only use of AES-CBC is in the static mode.

reynir: I added documentation in the README as well as in the code. While writing this I realized it should perhaps be:

```suggestion
        | B (Auth, _) -> Conf_map.mem Tls_auth t
```

As the `--auth` option is used for 1) static clients, 2) tls-auth; and for static clients we don't send connect options.

hannesm: This looks fine, but isn't `Secret` the same? I mean, wasn't `Secret` an alias for `Tls_auth`?

hannesm: ah no, nevermind.

src/config.ml
```diff
@@ -2278,7 +2278,16 @@ let client_generate_connect_options t =
...
         | B (Pull, _) -> true
         | B (Tls_mode, `Client) -> true
-        | B (Auth, _) -> true
+        | B (Auth, _) -> Conf_map.mem Tls_auth t
+            (* Assumption: we only support AEAD ciphers in tls clients. Thus
+               [Auth] option is ignored for the data channel, and for
+               tls-crypt(-v2) the hmac is hardcoded. For tls-auth we still want
+               to send it because it is used for authenticating the control
+               channel *)
+            not
```
hannesm, line 2287:
```suggestion
```

src/config.ml
```diff
@@ -2278,7 +2278,16 @@ let client_generate_connect_options t =
...
+              (Conf_map.mem Tls_crypt t
```
hannesm, line 2288:
```suggestion
```

src/config.ml
```diff
@@ -2278,7 +2278,16 @@ let client_generate_connect_options t =
...
+              || Conf_map.mem Tls_crypt_v2_client t
```
hannesm, line 2289:
```suggestion
```

src/config.ml
```diff
@@ -2278,7 +2278,16 @@ let client_generate_connect_options t =
...
+              || Conf_map.mem Tls_crypt_v2_server t)
```
hannesm, line 2290:
```suggestion
```

Issue #207, opened 2024-01-22
Handle the `key-direction` directive


dinosaure:
Currently, we must use:
```
tls-auth [inline] {0,1}
<tls-auth>
...
</tls-auth>
```

But some configurations use `key-direction`.

reynir:
This was fixed in #208 I believe.

PR #208, opened 2024-01-24
Be able to parse key-direction in our configuration file

dinosaure:


dinosaure:
Fix #207

reynir (commented):
I'm curious if this is a valid configuration:
```
key-direction 1
tls-auth [inline]
<tls-auth>
...
</tls-auth>
```
And also with `key-direction` and `tls-auth [inline]` swapped.

other than that I think it looks fine.

dinosaure:
Actually, if you put `tls-auth [inline]`, it leads to an error because we expect 2 inline parameters. Not sure how to fix it and if we should do.

reynir (approved):
It improves the situation even if there is an unhandled corner case. If you could `dune build @fmt` then I think it's good to merge.

test/config_tests.ml
```diff
@@ -381,6 +381,7 @@ let key_direction () =
     Fmt.str
       {|%a
 key-direction 1
+tls-auth [inline]
```
reynir, line 384: Did you mean to commit this line? As I understand your previous comment this makes the test fail.

dinosaure: Ah yes, I did the test locally and forget to remove it, sorry.

PR #210, opened 2024-02-22
Reduce allocations in data path

reynir:


reynir:
On a not-so-rigorous test this seems to increase download speeds from ~70-75 Mb/s to ~100-110 Mb/s. Openvpn achieves ~130 Mb/s but uses less CPU (about 50%).

reynir:
Please ignore the above. There is too much variation on my network to make this conclusion. Besides, these changes should mainly affect upload speeds(?)

src/engine.ml
```diff
@@ -1042,20 +1033,39 @@ let out ?add_timestamp (ctx : keys) hmac_algorithm compress rng data =
...
-    else Cstruct.empty
+  let aead (type key)
+      (authenticate_encrypt_tag :
+         key: key -> nonce: Cstruct.t -> ?adata: Cstruct.t -> Cstruct.t -> Cstruct.t * Cstruct.t)
+      (my_key : key)
+      my_implicit_iv
+      =
+      let nonce, replay_id =
+        let b = Cstruct.create_unsafe (Packet.id_len + Cstruct.length my_implicit_iv) in
+        Cstruct.BE.set_uint32 b 0 ctx.my_replay_id;
```
hannesm, line 1047:
```suggestion
        set_replay_id b 0;
```

src/engine.ml
```diff
@@ -1042,20 +1033,39 @@ let out ?add_timestamp (ctx : keys) hmac_algorithm compress rng data =
...
+      in
+      let enc, tag =
+        authenticate_encrypt_tag
+          ~key:my_key ~nonce ~adata:replay_id data
+      in
+      let res = Cstruct.create (4 + Cstruct.length tag + Cstruct.length enc) in
+      set_replay_id res 0;
+      Cstruct.blit tag 0 res 4 (Cstruct.length tag);
+      Cstruct.blit enc 0 res (4 + Cstruct.length tag) (Cstruct.length enc);
+      res
```
hannesm, line 1067:
```suggestion
      Cstruct.concat [ replay_id ; tag ; enc ]
```

reynir: Ah yes, I initially removed `replay_id` but now we have it again (throug reuse). Thanks!

hannesm (approved):
LGTM, please run an autoformat (`dune build @fmt --auto`) before merging.

reynir:
In https://github.com/reynir/miragevpn/tree/bench I implemented benchmarking for sending data in a (tls-auth) client.

Before rebasing on main (4db9951d7bdaaff0afdbbbed8915d9c2c4e3620d and thus this branch):
```
╭─────────────────┬───────────────────────────┬───────────────────────────┬───────────────────────────╮
│name             │  major-allocated          │  minor-allocated          │  monotonic-clock          │
├─────────────────┼───────────────────────────┼───────────────────────────┼───────────────────────────┤
│  Client encode  │             0.0000 mjw/run│           440.4918 mnw/run│           5955.9876 ns/run│
╰─────────────────┴───────────────────────────┴───────────────────────────┴───────────────────────────╯
```
And after rebasing on main:
```
╭─────────────────┬───────────────────────────┬───────────────────────────┬───────────────────────────╮
│name             │  major-allocated          │  minor-allocated          │  monotonic-clock          │
├─────────────────┼───────────────────────────┼───────────────────────────┼───────────────────────────┤
│  Client encode  │             0.0000 mjw/run│           377.5155 mnw/run│           5341.1295 ns/run│
╰─────────────────┴───────────────────────────┴───────────────────────────┴───────────────────────────╯
```

So the speedup is a more modest ~10% and ~15% less allocations. Not too bad, even if not 40% :)

hannesm:
Hmm, major-allocated and minor-allocated sound like the OCaml GC is measured? While this PR shuffles around Cstruct.t allocations (i.e. bigarray --> `malloc` on the external heap). Are they measured by your benchmark?


And I don't quite understand your numbers, the first being "Before rebasing on main (https://github.com/robur-coop/miragevpn/commit/4db9951d7bdaaff0afdbbbed8915d9c2c4e3620d and thus this branch):" -- does this include this PR or not? (I assume no)

"And after rebasing on main:" -- which main did you rebase what onto (I suspect your bench branch onto main including this PR merged into main)?

reynir:
OK let me rephrase:

Let main refer to the commit https://github.com/robur-coop/miragevpn/commit/4db9951d7bdaaff0afdbbbed8915d9c2c4e3620d. The first table is before merging main into the bench branch (I unfortunately didn't record what commit that was and I have since pushed more changes to the branch). The second table is after merging with main and includes the changes in this PR.

Looking at the bechamel documentation I think the bigarray allocations are **not** measured. Thus it is not very useful to draw conclusions from the allocation numbers (sorry).

From a manual review I counted cstruct allocations in `Engine.out` going from 11(12) for AES and 6(7) for AEAD to 5 and 4(5) cstruct allocations. Numbers in parenthesis is when `compression` is true. Many of the allocations removed are smaller allocations. The number of allocations could be reduced further if we could pass a destination buffer to the rng, hmac and the encrypt/authenticate_encrypt_tag operations

dinosaure:
> Looking at the bechamel documentation I think the bigarray allocations are not measured. Thus it is not very useful to draw conclusions from the allocation numbers (sorry).

Yes, that's true. This code show that `Gc.stats` (which is used by `bechamel`) does not count `malloc()`:
```ocaml
let () =
  let { Gc.major_words; _ } = Gc.quick_stat () in
  Format.printf "%fMw\n%!" major_words;
  let v = Cstruct.create 0x1000 in
  let { Gc.major_words; _ } = Gc.quick_stat () in
  Format.printf "%fMw\n%!" major_words
```

```sh
$ ocamlfind opt -linkpkg -package cstruct main.ml
$ ./a.out
97.000000Mw
97.000000Mw
```

Also, you actually can add a new metric with `bechamel` if you want which can count `malloc()` (of course, it will be a bit complex but that the objective of `bechamel`, to extend your benchmark if some system-wise/specific metrics).

PR #212, opened 2024-02-23
decode_tls_data: don't add u_len twice

reynir:
The value `u_len` is already added inside `p_start`. This check often passed due to the peer-info at the end; so happy coincidence.


hannesm (approved):
LGTM, nice find

PR #213, opened 2024-03-01
Avoid copying buffer when linger is empty

reynir:
`Cstruct.append Cstruct.empty buf` results in a _copy_ of `buf`. This is wasteful, so let's avoid it.

The performance difference is barely measurable, at least on my noisy laptop.


reynir:
Closing in favor of #214.

PR #214, opened 2024-03-01
More garbage friendly cstruct wrapper

reynir:
This supercedes #213. This adds a module `Cstruct_ext` with functions `append_nocopy` that does not copy if either argument is empty, and `sub` and `shift` variants that return `Cstruct.empty` instead of the result of `Cstruct.sub` or `Cstruct.shift` if the result is empty. The reason for this is an empty cstruct still keeps a live reference to its underlying bigarray.

### DIFF
--- a/src/cstruct_ext.ml
+++ b/src/cstruct_ext.ml
@@ -1,0 +1,20 @@
+(* [or_empty cs] is [cs] unless [Cstruct.is_empty cs] where it is [Cstruct.empty].
+   An empty cstruct will keep a live reference to its underlying buffer. It is
+   preferable to keep a reference live to [Cstruct.empty.buffer] than
+   [cs.buffer] if [cs] is empty. *)
+let[@ocaml.inline always] or_empty cs =
+  if Cstruct.is_empty cs then Cstruct.empty else cs
+
+(* [Cstruct.sub cs 0 cs.len] will result in an empty cstruct **that keeps a
+   live reference to [cs.buffer]**!! *)
+let sub cs off len = or_empty (Cstruct.sub cs off len)
+
+(* [Cstruct.shift cs cs.len] has a similar story *)
+let shift cs len = or_empty (Cstruct.shift cs len)
+
+(* If we append the empty cstruct and we don't need a fresh copy we can do
+   nothing. This has different semantics than Cstruct.append. *)
+let append_nocopy cs cs' =
+  if Cstruct.is_empty cs then cs'
+  else if Cstruct.is_empty cs' then cs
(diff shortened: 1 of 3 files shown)


src/cstruct_ext.ml
```diff
@@ -0,0 +1,20 @@
...
+(* [Cstruct.sub cs 0 cs.len] will result in an empty cstruct **that keeps a
+   live reference to [cs.buffer]**!! *)
+let sub cs off len = or_empty (Cstruct.sub cs off len)
+
+(* [Cstruct.shift cs cs.len] has a similar story *)
+let shift cs len = or_empty (Cstruct.shift cs len)
+
+(* If we append the empty cstruct and we don't need a fresh copy we can do
+   nothing. This has different semantics than Cstruct.append. *)
+let append_nocopy cs cs' =
```
hannesm, line 17: hmm hmm, the name is a bit misleading

src/cstruct_ext.ml
```diff
@@ -0,0 +1,20 @@
+(* [or_empty cs] is [cs] unless [Cstruct.is_empty cs] where it is [Cstruct.empty].
+   An empty cstruct will keep a live reference to its underlying buffer. It is
+   preferable to keep a reference live to [Cstruct.empty.buffer] than
+   [cs.buffer] if [cs] is empty. *)
+let[@ocaml.inline always] or_empty cs =
+  if Cstruct.is_empty cs then Cstruct.empty else cs
+
+(* [Cstruct.sub cs 0 cs.len] will result in an empty cstruct **that keeps a
+   live reference to [cs.buffer]**!! *)
+let sub cs off len = or_empty (Cstruct.sub cs off len)
```
hannesm, line 10: this hides potential errors, no? if I do a "Cstruct.sub Cstruct.empty 0 10", now I get the Cstruct.empty back -- previously I got an exception (not sure whether this is relevant or not).

src/cstruct_ext.ml
reynir: Right, it can't avoid copying in all cases. How about `append'` then?

hannesm: Yes, `append'` sounds good - or "append_avoid_copy_if_empty"? I'd be interested to upstream this into the Cstruct library, but then it modifies the semantics (that some may rely on, esp. in a concurrent setting), and also I try to remove Cstruct from packages ;)

reynir: I think they might be interesting to upstream. The `append` alternative would have to have a different name due to differing semantics that might matter in some cases. Then again as you say we probably want to use cstruct less.

PR #217, opened 2024-03-04
encode: reduce allocations

reynir:
A buffer that can hold the whole packet is allocated early on. Instead of reconstructing most of the packet in order to compute the signature it is computed on sub buffers of the packet.

Still TODO is:

- [x] tls-crypt encode
- [x] avoid having to allocate a dummy hmac just to pass the hmac size
- [x] the hmac dance could probably be simplified

### DIFF
--- a/src/engine.ml
+++ b/src/engine.ml
@@ -104,10 +104,14 @@ let compute_hmac key p hmac_algorithm hmac_key =
 
 let hmac_and_out protocol { hmac_algorithm; my_hmac; _ } key
     (p : [< Packet.ack | Packet.control ]) =
-  let hmac = compute_hmac key p hmac_algorithm my_hmac in
+  (* A dummy hmac. Unfortunately, it has to be the correct length for now *)
+  let hmac = Cstruct.create (Mirage_crypto.Hash.digest_size hmac_algorithm) in
   let header = Packet.header p in
   let p' = Packet.with_header { header with Packet.hmac } p in
-  Packet.encode protocol (key, p')
+  let buf, feeder = Packet.encode protocol (key, p') in
+  let hmac = Mirage_crypto.Hash.maci hmac_algorithm ~key:my_hmac feeder in
+  Packet.set_hmac buf protocol hmac;
+  buf
 
 let encrypt_and_out protocol { my; _ } key
     (p : [< Packet.ack | Packet.control ]) =
@@ -1116,7 +1120,7 @@ let data_out ?add_timestamp (ctx : keys) hmac_algorithm compress protocol rng
     key data =
   (* as described in [out], ~add_timestamp is only used in static key mode *)
   let ctx, payload = out ?add_timestamp ctx hmac_algorithm compress rng data in
-  let out = Packet.encode protocol (key, `Data payload) in
+  let out = Packet.encode_data protocol key payload in
   Log.debug (fun m ->
       m "sending %d bytes data (enc %d) out id %lu" (Cstruct.length data)
         (Cstruct.length out) ctx.my_replay_id);
--- a/src/engine.ml
+++ b/src/engine.ml
@@ -115,13 +115,13 @@ let encrypt_and_out protocol { my; _ } key
     (p : [< Packet.ack | Packet.control ]) =
   let my_key = Tls_crypt.Key.cipher_key my in
   let my_hmac = Tls_crypt.Key.hmac my in
-  let to_be_signed = Packet.Tls_crypt.to_be_signed key p in
-  let hmac = Mirage_crypto.Hash.SHA256.hmac ~key:my_hmac to_be_signed in
+  let buf, enc_off, enc_len, feeder =
+    Packet.Tls_crypt.encode protocol (key, p)
+  in
+  let hmac = Mirage_crypto.Hash.SHA256.hmaci ~key:my_hmac feeder in
   let iv = Cstruct.sub hmac 0 16 in
   let ctr = Mirage_crypto.Cipher_block.AES.CTR.ctr_of_cstruct iv in
-  let header = Packet.header p in
-  let p = Packet.with_header { header with Packet.hmac } p in
-  let buf, enc_off, enc_len = Packet.Tls_crypt.encode protocol (key, p) in
+  Packet.Tls_crypt.set_hmac buf protocol hmac;
   let encrypted =
     Mirage_crypto.Cipher_block.AES.CTR.encrypt ~key:my_key ~ctr
       (Cstruct_ext.sub buf enc_off enc_len)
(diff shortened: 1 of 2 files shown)


reynir:
I did similar changes for tls-crypt, but they are untested (for now)

src/engine.ml
```diff
@@ -104,22 +104,24 @@ let compute_hmac key p hmac_algorithm hmac_key =
 
 let hmac_and_out protocol { hmac_algorithm; my_hmac; _ } key
     (p : [< Packet.ack | Packet.control ]) =
-  let hmac = compute_hmac key p hmac_algorithm my_hmac in
-  let header = Packet.header p in
-  let p' = Packet.with_header { header with Packet.hmac } p in
-  Packet.encode protocol (key, p')
+  (* A dummy hmac. Unfortunately, it has to be the correct length for now *)
```
hannesm, line 107: I don't quite understand this comment...

reynir: Ah, it's no longer relevant since c635ce5576677614595ee3bc9343d5b426ed66e0. Before that a `Cstruct.create hmac_len` was added in the `p` header.

reynir: To expand a bit more: `Packet.encode` figured out the hmac length from the `hmac` cstruct member of the header. This design required passing a dummy hmac value of the correct length which was wasteful. Then I refactored so the int `hmac_len` is passed to `Packet.encode` and the `hmac` member is ignored (=> we can use `Cstruct.empty`!). Then the correct hmac is set using `Packet.set_hmac`.

src/packet.ml
```diff
@@ -239,6 +238,17 @@ let op_key op key =
...
 
+let protocol_len = function `Tcp -> 2 | `Udp -> 0
+
+let set_protocol buf proto =
+  match proto with
+  | `Tcp -> Cstruct.BE.set_uint16 buf 0 (Cstruct.length buf - 2)
+  | `Udp -> ()
+
+let set_hmac buf proto hmac =
+  let off = protocol_len proto + 1 + 8 in
```
hannesm, line 249: I'd appreciate if we can document where 1 and 8 comes from

src/packet.ml
```diff
@@ -247,21 +257,48 @@ let encode_protocol proto len =
...
   | `Udp -> Cstruct.empty
 
-let encode proto (key, p) =
-  let payload, len =
+let encode proto hmac_len
+    (key, (p : [< `Ack of header | `Control of operation * _ ])) =
+  let hdr = header p in
+  let len =
+    let id_arr_len = id_len * List.length hdr.ack_sequence_numbers in
+    protocol_len proto + 1 + hdr_len hmac_len + id_arr_len
```
hannesm, line 265: the 1 is the operation&key byte, isn't it?

src/packet.ml
```diff
@@ -247,21 +257,48 @@ let encode_protocol proto len =
...
+    + (if id_arr_len = 0 then 0 else 8)
```
hannesm, line 266: this is for some session id or so?

src/packet.ml
```diff
@@ -247,21 +257,48 @@ let encode_protocol proto len =
...
+    +
     match p with
-    | `Ack ack -> encode_header ack
-    | `Control (_, control) -> encode_control control
-    | `Data d -> (d, Cstruct.length d)
+    | `Ack _ -> 0
+    | `Control (_, (_, _, payload)) -> 4 + Cstruct.length payload
```
hannesm, line 270: the 4 would be nice to document it's whereabouts

src/packet.ml
```diff
@@ -247,21 +257,48 @@ let encode_protocol proto len =
...
+    match p with
+    | `Ack ack -> ignore (encode_header hmac_len to_encode ack)
+    | `Control (_, control) -> encode_control hmac_len to_encode control
+  in
+  let feeder feed =
+    (* replay_id ++ timestamp *)
+    feed (Cstruct.sub to_encode (hmac_len + 8) (4 + 4));
+    feed op_buf;
+    (* local_session *)
+    feed (Cstruct.sub to_encode 0 8);
```
hannesm, line 288: the binding to op_buf can be removed.

We can `feed (Cstruct.sub buf (protocol_len proto) 9)` -- which includes the op_buf and the local session then :)

src/packet.ml
```diff
@@ -346,42 +385,50 @@ module Tls_crypt = struct
...
-      | `Control (op, control) -> encode_control op control
+      | `Control (Hard_reset_client_v3, (_, _, wkc)) -> Cstruct.length wkc
+      | _ -> 0
     in
-    let op_buf =
-      let b = Cstruct.create 1 in
-      let op = op_key (operation p) key in
-      Cstruct.set_uint8 b 0 op;
-      b
+    let buf = Cstruct.create len in
```
hannesm, line 411: further above, you used create_unsafe -- we may use that here as well?

hannesm:
this looks great! some minor remarks, and one train of thought is: should we use create_unsafe since we fill the entire buffer anyways?

src/packet.ml
reynir: Yes, correct

hannesm: maybe add a comment behind the number so we don't have to guess in the future? ;)

src/packet.ml
reynir: I pushed a commit that uses `create_unsafe`. I chose not to use `create_unsafe` at first because allocation and filling were done in different parts of the code. But it should be safe!

hannesm: :D

src/engine.ml
```diff
@@ -1116,7 +1117,7 @@ let data_out ?add_timestamp (ctx : keys) hmac_algorithm compress protocol rng
     key data =
   (* as described in [out], ~add_timestamp is only used in static key mode *)
   let ctx, payload = out ?add_timestamp ctx hmac_algorithm compress rng data in
-  let out = Packet.encode protocol (key, `Data payload) in
+  let out = Packet.encode_data protocol key payload in
```
hannesm, line 1120: this I guess is the path that we want to optimize the most.

right now, `outgoing` (below) is called with the data-to-be-sent. This calls `static_out` or `data_out`, which both call `out` followed by `encode_data` (or `encode_protocol`).

I'd like that `out` (which does a `Cstruct.concat`) already returns the (potentially) required prefix -- so in `encode_data` we won't need another allocation and blit. WDYT?

reynir:
It would be great to make similar changes for decode regarding hmac. That is, compute the hmac on the input buffer directly instead of reconstructing it from the parsed data. However, the decoding is done in parts and we "lose" the op_key part early on. We can reduce allocations in `to_be_signed` a bit, but ideally we would remove it.

src/engine.ml
reynir: Yes, sounds good! I'll take a look.

hannesm: I briefly looked into `out`, and came up with #218 -- may be worth to merge before modifying it further.

hannesm:
We can for sure do that for decoding :) I also agree that decoding and reassembling / reconstructing it just for the hmac isn't a great idea.

hannesm (approved):
LGTM

hannesm:
this is a nice improvement. Feel free to merge. any further improvements can be done in separate PRs :)

PR #219, opened 2024-03-05
Data packet encode allocations

reynir:
A continuation of #217.

### DIFF
--- a/src/cstruct_ext.ml
+++ b/src/cstruct_ext.ml
@@ -18,3 +18,17 @@ let append' cs cs' =
   if Cstruct.is_empty cs then cs'
   else if Cstruct.is_empty cs' then cs
   else Cstruct.append cs cs'
+
+let concat_with_empty_prefix prefix_len css =
+  let len = prefix_len + Cstruct.lenv css in
+  let b = Cstruct.create_unsafe len in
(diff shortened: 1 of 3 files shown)


src/cstruct_ext.ml
```diff
@@ -18,3 +18,17 @@ let append' cs cs' =
   if Cstruct.is_empty cs then cs'
   else if Cstruct.is_empty cs' then cs
   else Cstruct.append cs cs'
+
+let concat_with_empty_prefix prefix_len css =
+  let len = prefix_len + Cstruct.lenv css in
+  let b = Cstruct.create_unsafe len in
+  Cstruct.memset (Cstruct.sub b 0 prefix_len) 0;
```
hannesm, line 25: I'd drop this memset. we especially create these bytes as slack space to fill them thereafter.

hannesm (approved):
LGTM

src/cstruct_ext.ml
reynir: Hmm, maybe if we rename it `concat_with_unsafe_prefix`? It is about memset'ing 0-3 bytes so maybe not the biggest performance worry. WDYT?

hannesm: sounds good to me, the renaming :) not calling memset is likely better than calling it, performance-wise.

PR #224, opened 2024-04-11
Refactor common bindings to app/common (not exposed)

hannesm:


reynir (approved):
Thanks! This is quite nice.

PR #225, opened 2024-04-11
add a lwt miragevpn_server_notun to ease development

hannesm:


app/miragevpn_server_notun.ml
```diff
@@ -0,0 +1,255 @@
...
+  (* foreach connection, call handle `Tick and follow instructions! *)
+  IPM.fold
+    (fun k (fd, t) acc ->
+      acc >>= fun acc ->
+      match Miragevpn.handle !t `Tick with
+      | Error e ->
+          Logs.err (fun m -> m "error in timer %a" Miragevpn.pp_error e);
+          Lwt.return acc
+      | Ok (_t', _out, _payloads, Some `Exit) ->
+          (* TODO anything to do with "_out" or "_payloads"? *)
```
reynir, line 84: I don't think we can produce any `payload`s this way. But we might produce `out`s - e.g. retransmission. Then looking at engine.ml we don't implement that yet in the server.

app/miragevpn_server_notun.ml
```diff
@@ -0,0 +1,255 @@
...
+          Logs.warn (fun m -> m "exiting %a" Ipaddr.V4.pp k);
+          Lwt.return acc
+      | Ok (t', out, payloads, act) -> (
+          (* TODO anything to do with "_act"? (apart from exit) *)
```
reynir, line 88: Here it could be nice if the type system would help us comb out the client specific actions :/ but indeed I think we only need to handle `` `Exit ``.

app/miragevpn_server_notun.ml
```diff
@@ -0,0 +1,255 @@
...
+          Option.iter
+            (fun a ->
+              Logs.warn (fun m ->
+                  m "in timer, ignoring action %a" Miragevpn.pp_action a))
+            act;
+          t := t';
+          Lwt_list.iter_p (handle_payload server k) payloads >>= fun () ->
```
reynir, line 95: I don't think we can produce payloads due to `` `Tick `` event. But it doesn't hurt (much) to iterate over the empty list.

app/miragevpn_server_notun.ml
reynir: Ah I see this comment is due to the `` `Exit `` action.

reynir (approved):
Nice! I tested it with our tlsauth-user setup, but it fails due to tls-mode not being explicitly present (it is implicitly present due to the `server` option). But the code looks good.

app/miragevpn_server_notun.ml
```diff
@@ -0,0 +1,262 @@
...
+  let open Lwt.Infix in
+  let is_not_taken ip = not (IPM.mem ip t.connections) in
+  let client_state = ref (Miragevpn.new_connection t.server) in
+  let dst =
+    match Lwt_unix.getsockname fd with
+    | Lwt_unix.ADDR_UNIX _ -> assert false
+    | Lwt_unix.ADDR_INET (ip, port) -> (
+        match Ipaddr_unix.V4.of_inet_addr ip with
+        | Some ip -> (ip, port)
+        | None -> assert false)
```
reynir, line 144: This fails because we listen on ipv6 :/

reynir:
There is some confusion about the VPN layer ip address and the network layer ip address :/

reynir:
Ugh I accidentally committed the bench/ directory :/

app/miragevpn_server_notun.ml
```diff
@@ -0,0 +1,255 @@
...
+      Logs.err (fun m -> m "destination %a not found in map" Ipaddr.V4.pp dst);
+      Lwt.return_unit
+  | Some (fd, state) -> (
+      match Miragevpn.outgoing !state cs with
+      | Error `Not_ready ->
+          Logs.err (fun m ->
+              m "error not_ready while writing to %a" Ipaddr.V4.pp dst);
+          Lwt.return_unit
+      | Ok (state', enc) -> (
+          (* TODO fragmentation!? *)
```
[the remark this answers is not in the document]
hannesm: (copied from mirage/miragevpn_mirage Server functor)... not sure, but not relevant in the notun case in any case :)

app/miragevpn_server_notun.ml
hannesm: yes, can we end up in exit and should we notify the peer? I think newer openvpn had something along the ride

app/miragevpn_server_notun.ml
hannesm: tragic, we should fix our code :)

hannesm:
Thanks for your fixes & merge :)

app/miragevpn_server_notun.ml
reynir: hmm I think there is a mechanism to notify-on-exit, but we don't implement that (yet) as far as I know. But I now realize we don't do anything **here** on `` `Exit ``, for example closing the file descriptor :/

PR #228, opened 2024-04-14
adapt mirage/miragevpn_mirage to changes in miragevpn_server_notun from @reynir

hannesm:
thanks for your fixes @reynir  -- I manually pushed them to the mirage server implementation :)


reynir:
Ah, thanks for adding the comment in miragevpn_server_notun -- I did not realize where the code came from.

PR #163, opened 2023-11-16
Support TLS-EKM

hannesm:
needs https://github.com/mirleft/ocaml-tls/pull/482
fixes #103 

also pushed IV_PROTO into a bitmap and types (see Packet.Iv_proto)

//cc @reynir

PR #232, opened 2024-04-18
Keep TLS state when established

reynir:
We currently will not use the TLS state once established, but if we implement exit control messages we will need to keep the TLS state around.

We may not need the TLS state if exitcc is not negotiated. In that case we can reiterate and make it an option.

### DIFF
--- a/src/engine.ml
+++ b/src/engine.ml
@@ -614,7 +614,7 @@ let kex_server config session (my_key_material : my_key_material) tls data =
           kdf ~tls_ekm session cipher hmac_algorithm my_key_material
             their_tls_data
         in
-        Ok (`State (Established keys_ctx, Some ip_config))
+        Ok (`State (Established (tls', keys_ctx), Some ip_config))
     | _ ->
         Error (`Msg "found Ifconfig without IPv4 addresses, not yet supported")
   in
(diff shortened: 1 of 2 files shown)


src/engine.ml
```diff
@@ -808,15 +808,15 @@ let incoming_control_client config rng session channel now op data =
   | Push_request_sent (tls, key, tls_data), Packet.Control ->
       let open Result.Syntax in
       Log.debug (fun m -> m "in push request sent");
-      let* _tls', d = incoming_tls_without_reply tls data in
+      let* tls', d = incoming_tls_without_reply tls data in
       let+ config' = maybe_push_reply config d in
       let cipher = Config.get Cipher config'
       and hmac_algorithm = Config.get Auth config'
       and tls_ekm =
         Option.map (fun `Tls_ekm -> tls) (Config.find Key_derivation config')
```
reynir, line 816: Here I always wonder why we use `tls` and not `tls'`. Presumably it doesn't matter which one we use as long as the tls handshake is finished?!

hannesm: yes, well, best to use` tls'` here :) maybe that explains the issues on some debian systems (ref #163)

hannesm:
looks good to me, thanks!

PR #234, opened 2024-04-19
When `Established _` read incoming control channel messages

reynir:
When the openvpn tunnel is established keep reading control channel messages. For now we just warn about unknown control messages and do nothing. The [OpenVPN implementation](https://github.com/OpenVPN/openvpn/blob/32e6586687a548174b88b64fe54bfae6c74d4c19/src/openvpn/forward.c#L233-L301) also just warns for unknown control messages it seems. I may extend this PR to handle exit and restart messages. 

The `EXIT` and `RESTART` control channel messages are only emitted by OpenVPN when using UDP. Therefore I have only tested it with our miragevpn-client-notun connecting to an OpenVPN server with `--explicit-exit-notify`. Restarting the OpenVPN service resulted in a warning about an unknown control channel message `"RESTART\000"` and the client kept trying to send pings in the VPN. This is different than from before where it would fail with "no transition found". I am not sure this is actually desirable...

### DIFF
--- a/src/engine.ml
+++ b/src/engine.ml
@@ -930,7 +930,7 @@ let incoming_control_server is_not_taken config rng session channel _now _ts
         | None -> `Reset_server
         | Some _ -> `Reset
       in
-      Ok (None, config, session, channel, [ (control_typ, Cstruct.empty) ])
+      Ok (None, config, channel, [ (control_typ, Cstruct.empty) ])
   | TLS_handshake tls, Packet.Control -> (
       let open Result.Syntax in
       let* tls', tls_response, d = incoming_tls tls data in
@@ -954,9 +954,8 @@ let incoming_control_server is_not_taken config rng session channel _now _ts
           let* ip_config, config, channel_st, out' =
             server_handle_tls_data config is_not_taken session keys tls' d
           in
-          Ok
-            (ip_config, config, session, { channel with channel_st }, out @ out')
-      | _ -> Ok (None, config, session, { channel with channel_st }, out))
+          Ok (ip_config, config, { channel with channel_st }, out @ out')
+      | _ -> Ok (None, config, { channel with channel_st }, out))
   | TLS_established (tls, keys), Packet.Control -> (
       let open Result.Syntax in
       let* tls', d = incoming_tls_without_reply tls data in
@@ -965,10 +964,10 @@ let incoming_control_server is_not_taken config rng session channel _now _ts
           let* ip_config, config, channel_st, out =
             server_handle_tls_data config is_not_taken session keys tls' d
           in
-          Ok (ip_config, config, session, { channel with channel_st }, out)
+          Ok (ip_config, config, { channel with channel_st }, out)
       | None ->
           let channel_st = TLS_established (tls', keys) in
-          Ok (None, config, session, { channel with channel_st }, []))
+          Ok (None, config, { channel with channel_st }, []))
   | Push_request_sent (tls, key, tls_data), Packet.Control ->
       (* TODO naming: this is actually server_stuff sent, awaiting push request *)
       let open Result.Syntax in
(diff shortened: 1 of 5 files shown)


reynir:
I pushed two commits that fix reading end-of-file in miragevpn-client-notun using TCP, and another that fixes `--connect-retry-max` logic.

I have discovered the management interface in OpenVPN where the server (or client) listens on a socket and accepts some simple commands. With that I can send a command `client-kill CID [message]` where `CID` is the client id which is a number starting from 0 increasing for every new client connection. The optional message will replace `"RESTART\000"` with the message terminated by a NUL byte. After sending the message the server will wait about five seconds and then close the socket (if using TCP).

What is interesting to me is that we **don't** advertice exit-cc and yet the OpenVPN server sends a `RESTART` message to our client. And it does so even we are connecting using TCP.

reynir:
There are some commits that are "off topic" and fixes some bugs or refactors the code - those could be cherry-picked.

I added a function `send_control_message` to `Miragevpn.mli` which allows the caller to send arbitrary control channel messages. We may not want to expose it directly (I don't know), but we probably want to at least expose a way to send `EXIT`, `RESTART` and such.

I also made the client and the server `handle` return `` `Exit `` on `EXIT`, `HALT` and `RESTART` control channel messages. This is not correct. I think for `RESTART` we would like to reconnect either to the same remote or the next in the list.

The state on reconnect would also have to be reset. I notice that we can reconnect in some cases, but the logic ignores `PUSH_REPLY` because we received a `PUSH_REPLY` in the previous session and did not clean up the state properly.

During my testing I discovered an `EXIT` message makes the peer wait 5 seconds before disconnecting -- but if you keep sending `EXIT` messages fast enough the timer will reset and OpenVPN will never disconnect. I also learned that it is possible to send `INFO` messages to servers, and the server will confusingly print log messages saying it received a message from the server (even if it was from a client): `Info command was pushed by server ('You know the rules and so do I (do I)')`

src/engine.ml
```diff
@@ -1562,6 +1619,42 @@ let wrap_control state control_crypto needs_wkc key transport outs =
...
+  | #control_tls as control_crypto ->
+      let* channel_st, out =
+        match (s.state, s.channel.channel_st) with
+        | (Client Ready | Server Server_ready), Established (tls, keys) ->
+            let data' = Cstruct.create_unsafe (String.length data + 1) in
+            Cstruct.blit_from_string data 0 data' 0 (String.length data);
+            Cstruct.set_uint8 data' (String.length data) 0;
+            let* tls, out =
+              (* reynir: I *think* it only returns [None] when not established
+                 or after write is shutdown. *)
```
hannesm, line 1637: yes :) that assumption is shared with me -- the send_application_data shouldn't return None if the channel is ready.

reynir:
>     * what is the difference between halt and exit?
The openvpn-rfc has sections on [RESTART and HALT](https://github.com/OpenVPN/openvpn-rfc/blob/71d6b69005b223a395c65db89693e5d70abb0e48/openvpn-wire-protocol.xml#L1802-L1828) and [EXIT](https://github.com/OpenVPN/openvpn-rfc/blob/71d6b69005b223a395c65db89693e5d70abb0e48/openvpn-wire-protocol.xml#L1877-L1888).
It is rather unclear what the exact semantics are. The text on `RESTART` and `HALT` is garbled and confusing. In the code RESTART and HALT result in calling `server_pushed_signal()` and I think `RESTART` means reconnect to the server (or the next server depending on flags). I *think* `HALT` can make the openvpn peer exit. `server_pushed_signal()` checks for `--pull` so it should only work on clients.

`EXIT` results in a call to `receive_exit_message()` and works both on the client and the server. When a server received `EXIT` it waits 5 seconds and then closes the connection it seems (the 5 seconds are so an ACK for the EXIT has time to arrive; mostly relevant for UDP). For the client I think it will restart on `EXIT` immediately.

>     * should we already have an explicit `` `Restart `` return value?
We should probably handle RESTART, HALT and EXIT differently, yes.,

>     * can both the server and the client send these control messages?
It is valid to send all these control messages. For RESTART and HALT they will be silently ignored by the server, though. Freshly made up control channel messages can also be sent - an OpenVPN peer will just log a warning about unknown messages and nothing else.

Issue #181, opened 2023-12-04
IV_PROTO tls-ekm does not work nicely with a debian server


hannesm:
As remarked by @reynir in https://github.com/robur-coop/miragevpn/pull/169#issuecomment-1838147741:
I observed AEAD decrypt failures on both the OpenVPN™ server and the client when using tls 1.3 on my own debian server with both tls-crypt and tls-crypt-v2. Using tls 1.2 works fine however. I tested with the main branch and I can reproduce the same behavior (except for tls-crypt of course).

Interestingly I can't reproduce against the OpenVPN™ server running on FreeBSD.

@reynir: could you add OpenVPN version numbers to this issue so at a later point in time we can try to reproduce with the exact versions?

On the FreeBSD server, we use openvpn-2.6.8

reynir:
I used openvpn 2.6.3 and a custom build with extra debug logging and tls-crypt protocol_dump fix 2.7_git. In both cases OpenSSL 1.1.1w is used.

hannesm:
ah, good that you mention it -- we are using OpenSSL 1.1.1h-freebsd  22 Sep 2020 on the FreeBSD machine

hannesm:
With https://github.com/mirleft/ocaml-tls/pull/495 I can establish an EKM session with TLS 1.3 between our server and client (both notun).

The old Debian server side issue may be related to https://github.com/openssl/openssl/issues/3940 -- but I wasn;t able to track down which OpenSSL versions are affected.

I'd suggest: let's get a tls release out, and then enable EKM in MirageVPN (and if we encounter another issue, we can revert).

PR #235, opened 2024-04-24
Server tls-ekm negotiation

reynir:
This implements tls-ekm negotiation in the server. With this change I am able to reproduce #181 in the miragevpn-server-notun both with OpenVPN client and our Miragevpn client **when using tls 1.3**. When using tls 1.2 it works just fine with both clients.

### DIFF
--- a/src/config.ml
+++ b/src/config.ml
@@ -139,6 +139,20 @@ let topology_to_string = function
   | `P2p -> "p2p"
   | `Subnet -> "subnet"
 
+module Protocol_flag = struct
+  (* Known protocol flags *)
+  type t = [ `Cc_exit | `Tls_ekm ]
+
+  let to_string = function `Cc_exit -> "cc-exit" | `Tls_ekm -> "tls-ekm"
+
+  let of_string = function
+    | "cc-exit" -> Some `Cc_exit
+    | "tls-ekm" -> Some `Tls_ekm
+    | _ -> None
+
+  let compare : t -> t -> _ = compare
+end
+
 module Conf_map = struct
   type flag = unit
 
@@ -202,6 +216,7 @@ module Conf_map = struct
           * [ `Udp | `Tcp of [ `Server | `Client ] option ])
           k
       (* see socket.c:static const struct proto_names proto_names[] *)
+    | Protocol_flags : Protocol_flag.t list k
     | Remote
         : ([ `Domain of [ `host ] Domain_name.t * [ `Ipv4 | `Ipv6 | `Any ]
            | `Ip of Ipaddr.t ]
@@ -301,6 +316,10 @@ module Conf_map = struct
         (`Msg
           "While --ca and --peer-fingerprint are not mutually exclusive the \
            semantics are unclear to us and thus not implemented")
+    else if mem Key_derivation t then
+      Error (`Msg "The --key-derivation option is reserved for push replies")
+    else if mem Protocol_flags t then
+      Error (`Msg "The --protocol-flags option is reserved for push replies")
     else Ok ()
 
   let cert_key_or_pkcs12 t =
@@ -567,6 +586,12 @@ module Conf_map = struct
           | `Tcp (Some `Client) -> "-client"
           | `Tcp (Some `Server) -> "-server"
           | `Tcp None | `Udp -> "")
+    | Protocol_flags, flags ->
+        p () "protocol-flags%a"
+          Fmt.(
+            list ~sep:nop
+              (append (any " ") (of_to_string Protocol_flag.to_string)))
+          flags
     | Pull, () -> p () "pull"
     | Remote, lst ->
         p () "%a"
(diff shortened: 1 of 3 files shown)


src/miragevpn.mli
```diff
@@ -217,11 +217,6 @@ module Config : sig
   val eq : eq
   (** [eq] is an implementation of [cmp] for use with [{!equal} cmp t t2] *)
 
-  val client_generate_connect_options :
-    t -> (string, [> `Msg of string ]) result
-  (** Exports the excerpts from the client configuration sent to the server
-      when the client initially connects. *)
-
```
reynir, line 217: This function seems very internal to me and I don't know why it was exposed in the first place - so let's not expose it anymore!

src/config.ml
```diff
@@ -567,6 +586,11 @@ module Conf_map = struct
           | `Tcp (Some `Client) -> "-client"
           | `Tcp (Some `Server) -> "-server"
           | `Tcp None | `Udp -> "")
+    | Protocol_flags, flags ->
+        (* XXX: We assume [flags] is always non-empty *)
+        p () "protocol-flags %a"
+          Fmt.(list ~sep:(any " ") (of_to_string Protocol_flag.to_string))
+          flags
```
reynir, line 593: There are no guarantees flags is not empty. However, a trailing space is not the end of the world...

src/config.ml
```diff
@@ -202,6 +216,7 @@ module Conf_map = struct
           * [ `Udp | `Tcp of [ `Server | `Client ] option ])
           k
       (* see socket.c:static const struct proto_names proto_names[] *)
+    | Protocol_flags : Protocol_flag.t list k
```
reynir, line 219: I considered using a set, but a list is simpler and there's at most half a dozen different flags at the moment.

src/engine.ml
```diff
@@ -558,7 +571,34 @@ let kex_server config session (my_key_material : my_key_material) tls data =
...
+      if
+        supports_tls_ekm
+        && Option.fold iv_proto ~none:false
+             ~some:Packet.Iv_proto.(contains Use_cc_exit_notify)
+      then Config.add_protocol_flag `Tls_ekm config
+      else config
+    in
+    if supports_tls_ekm then Config.add Key_derivation `Tls_ekm config
+    else config
+  in
```
reynir, line 601: After testing it seems OpenVPN **only** sends `protocol-flags tls-ekm` and no `key-derivation tls-ekm` if `Use_cc_exit_notify` bit is set. I guess it could make the code slightly simpler only to add either.

reynir:
I patched OpenVPN to print whether tls-ekm or prf key derivation is used, and with TLS 1.3 tls-ekm is used - but somehow AEAD decrypt fails with tls 1.3.

hannesm:
Thanks for your work. To summarize:
- with tls-ekm and TLS 1.3, our server doesn't work with our client neither our server
- in #181 our client doesn't work with TLS 1.3 and ekm with an openvpn server

The "but somehow AEAD decrypt fails" is only with tls-ekm AFAIU.

Your other changes in here, would you mind to submit them as separate PR so we can merge them, and keep this for working/debugging on tls-ekm?

reynir:
I cleaned up the history including undoing sending `Tls_key_export` in the client. I also added a commit which effectively disables tls-ekm in the server while keeping the logic in a similar manner to how we did it in the client.

So I think this can be merged. Otherwise, I'm unsure what it is you want separated.

src/config.ml
```diff
@@ -1664,6 +1709,7 @@ let a_config_entry : line A.t =
          a_dhcp_option;
          a_proto;
          a_proto_force;
+         a_protocol_flags <?> "protocol-flags";
```
hannesm, line 1712: from what I remember, shouldn't we put the most eager at front -- i.e. don't let `a_proto` consume the "proto", but put the `a_protocol_flags` further above so it can try to consume the "protocol-flags"?

src/config.ml
```diff
@@ -2371,8 +2371,6 @@ let client_generate_connect_options t =
   (* Ok "V4,dev-type tun,link-mtu 1560,tun-mtu 1500,proto TCPv4_CLIENT,\
      keydir 1,cipher AES-256-CBC,auth SHA1,keysize 256,tls-auth,\
      key-method 2" *)
-  let open Result.Syntax in
-  let* () = Conf_map.is_valid_client_config t in
```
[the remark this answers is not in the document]
hannesm: sounds fine to me.

hannesm (approved):
looks fine to me, apart from the minor comment about the order in the parser.

src/config.ml
```diff
@@ -890,6 +919,22 @@ let a_proto_force =
   *> choice [ string "tcp" *> return `Tcp; string "udp" *> return `Udp ]
   >>| fun prot -> `Proto_force prot
 
+let a_protocol_flags =
+  string "protocol-flags" *> a_whitespace *> commit
+  *> sep_by a_whitespace a_single_param
```
reynir, line 924: Here we can parse an empty `protocol-flags` (which is fine but probably not used much). However, the empty `protocol-flags` requires at least one space or comment at the end to parse successfully. This is maybe not a big deal, but since the server may send such a protocol-flags it's unfortunate to error on that.

I'm not sure how to fix that.

reynir: Figured it out. Tested manually with Miragevpn.Config.{parse,pp} in a top level.

hannesm:
Thanks for your replies and additional work.

PR #239, opened 2024-04-26
Use key method mask

reynir:
OpenVPN only checks the lower half of the key method in a key exchange message. It is unclear why and how the upper half might be used. I have not found any evidence the upper half has ever been used for anything other than zero bits.

Relevant OpenVPN code:
- https://github.com/OpenVPN/openvpn/blob/32e6586687a548174b88b64fe54bfae6c74d4c19/src/openvpn/ssl.h#L115-L116
- https://github.com/OpenVPN/openvpn/blob/32e6586687a548174b88b64fe54bfae6c74d4c19/src/openvpn/ssl.c#L2194-L2196

### DIFF
--- a/src/packet.ml
+++ b/src/packet.ml
@@ -547,6 +547,9 @@ let pp_tls_data ppf t =
 
 let key_method = 0x02
 
+(* For some reason OpenVPN only looks at the lower half of the byte *)
+let key_method_mask = 0x0f
+
 (* strings are
    (a) length-prefixed (2 bytes, big endian);
    (b) terminated with 0 byte;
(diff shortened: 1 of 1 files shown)


reynir:
For now I am in favor of **not** merging this as it seems key method 2 does not use the upper half of the byte for anything. So the mask may have been there for a previous key method version which we don't care about - or for future versions which we don't support either.

reynir:
The openvpn-rfc does not mention anything about the mask so let's not do it.

https://github.com/OpenVPN/openvpn-rfc/blob/71d6b69005b223a395c65db89693e5d70abb0e48/openvpn-wire-protocol.xml#L1364

PR #242, opened 2024-05-13
Add tls-crypt for our server, require if tls-mode is present as well

hannesm:
tls-auth or tls-crypt or tls-crypt-v2

### DIFF
--- a/src/config.ml
+++ b/src/config.ml
@@ -302,7 +302,7 @@ module Conf_map = struct
       | Some (`Tun, _) | None -> Ok ()
       | Some (`Tap, _) -> Error (`Msg "'dev-type tap' is not supported")
     in
-    if not (mem Tls_mode t || mem Server t) then
+    if not (mem Tls_mode t) then
       let* () =
         Result.map_error
           (fun err -> `Msg ("not a valid config: " ^ err))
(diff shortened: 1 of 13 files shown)


src/config.ml
```diff
@@ -1692,6 +1717,9 @@ let a_not_implemented =
       string "explicit-exit-notify";
       string "tun-ipv6";
       string "ifconfig-ipv6";
+      string "max-clients";
+      string "setenv";
+      string "push";
```
reynir, line 1722: It was useful for testing to ignore `push` directives, but eventually we may want to support `push` at least to some degree.
```suggestion
```

src/config.ml
```diff
@@ -1692,6 +1717,9 @@ let a_not_implemented =
       string "explicit-exit-notify";
       string "tun-ipv6";
       string "ifconfig-ipv6";
+      string "max-clients";
+      string "setenv";
```
reynir, line 1721: This should be safe to ignore as we don't use it in Mirage at least.

src/config.ml
```diff
@@ -1683,6 +1706,8 @@ let a_not_implemented =
       string "status";
       string "log-append";
       string "log";
+      string "user";
+      string "group";
```
reynir, line 1710: Nonsense options when it comes to Mirage :D

src/config.ml
```diff
@@ -1692,6 +1717,9 @@ let a_not_implemented =
       string "explicit-exit-notify";
       string "tun-ipv6";
       string "ifconfig-ipv6";
+      string "max-clients";
```
reynir, line 1720: Could *maybe* be useful to support, but let's ignore it for now.

reynir (approved):

PR #243, opened 2024-05-14
Reenable tls-ekm

### LINKED ISSUE #181: IV_PROTO tls-ekm does not work nicely with a debian server

reynir:
The bug has been identified in ocaml-tls and fixed by @hannesm. Clients and servers should again use tls-ekm for key derivation if possible.

Closes #181.

### DIFF
--- a/miragevpn.opam
+++ b/miragevpn.opam
@@ -36,7 +36,7 @@ depends: [
   "mirage-crypto"
   "mirage-crypto-rng" {>= "0.11.0"}
   "ptime"
-  "tls" {>= "0.17.4"}
+  "tls" {>= "0.17.5"}
   "dns-client" {>= "7.0.0"}
   "x509" {>= "0.15.0"}
   "duration"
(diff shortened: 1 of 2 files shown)


hannesm:
looks good, wait for https://github.com/ocaml/opam-repository/pull/25870 and we should adjust the tls bound to >= 0.17.5 :)

hannesm:
thanks, CI is green and the tls release is merged. :)

PR #244, opened 2024-05-14
Server tls-crypt-v2 support

reynir:
Implement tls-crypt-v2 in the server allowing configurations with both tls-crypt-v2 and either tls-auth or tls-crypt.

### DIFF
--- a/src/engine.ml
+++ b/src/engine.ml
@@ -327,111 +327,6 @@ let server server_config ~is_not_taken server_ts server_now server_rng =
     server_ip server_config,
     port )
 
-let new_connection server data =
-  let open Result.Syntax in
-  (* XXX(reynir): for now only TCP server *)
-  let protocol = `Tcp in
-  let session =
-    init_session
-      ~my_session_id:(Randomconv.int64 server.server_rng)
-      ~protocol ()
-  in
-  let current_ts = server.server_ts () in
-  let channel = new_channel 0 current_ts in
-  match Config.find Tls_crypt_v2_server server.server_config with
-  | None ->
-      let+ control_crypto =
-        match
-          (tls_auth server.server_config, tls_crypt server.server_config)
-        with
-        | Ok auth, Error _ -> Ok (`Tls_auth auth)
-        | Error _, Ok crypt -> Ok (`Tls_crypt (crypt, None))
-        | _ -> Error (`Msg "server only supports tls-auth or tls-crypt")
-      in
-      {
-        config = server.server_config;
-        is_not_taken = server.is_not_taken;
-        control_crypto;
-        state = Server Server_handshaking;
-        linger = Cstruct.empty;
-        rng = server.server_rng;
-        ts = server.server_ts;
-        now = server.server_now;
-        session;
-        channel;
-        lame_duck = None;
-        last_received = current_ts;
-        last_sent = current_ts;
-      }
-  | Some (_wrapping_key, true) ->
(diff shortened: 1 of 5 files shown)


reynir (commented):
This is a first draft of tls-crypt-v2 support. I am not very pleased with it. `new_connection` now takes the first packet as input to determine if we are using tls-crypt-v2 or either tls-auth or tls-crypt in setups that allow both. The packet then has to be passed to `handle` to do remaining parsing. This is in my opinion poor ergonomic and I plan to experiment with a different interface.

src/engine.ml
```diff
@@ -324,34 +327,110 @@ let server server_config ~is_not_taken server_ts server_now server_rng =
     server_ip server_config,
     port )
 
-let new_connection server =
+let new_connection server data =
   let open Result.Syntax in
+  (* XXX(reynir): for now only TCP server *)
+  let protocol = `Tcp in
```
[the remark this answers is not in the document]
hannesm: fine with that. I'm not so convinced of the `XXX` comment -- it doesn't add much to the code. if we want to track that server is TCP only right now, I'd prefer to have a GitHub issue open (and tbh, I'd open it once there's need for server on UDP).

src/engine.ml
```diff
@@ -324,34 +327,110 @@ let server server_config ~is_not_taken server_ts server_now server_rng =
     server_ip server_config,
     port )
 
-let new_connection server =
+let new_connection server data =
   let open Result.Syntax in
+  (* XXX(reynir): for now only TCP server *)
```
reynir, line 332:
```suggestion
```

hannesm (approved):
Great! :)

hannesm:
I reviewed by avoiding the whitespace changes (adding `?w=1` to the url). This looks good to me.

src/engine.ml
```diff
@@ -1918,6 +1892,101 @@ let incoming state control_crypto buf =
...
+                  last_received = current_ts;
+                  last_sent = current_ts;
+                },
+                control_crypto,
+                actual_packet )
+        | Ok (_not_hard_reset_client, _key, _received, _linger) ->
+            Error (`Msg "invalid initial packet"))
+    (* `No_transition *)
+  in
+  incoming t control_crypto packet
```
hannesm, line 1988: in incoming, there's:
```OCaml
              | Packet.Hard_reset_client_v3 ->
                  ignore_udp_error (Error (`No_transition (ch, op, received)))
```

which we should remove now, shouldn't we? and the serve should then also in `incoming_control_server` deal with reset_3...

src/engine.ml
```diff
@@ -1918,6 +1892,101 @@ let incoming state control_crypto buf =
...
+            tls_auth_or_tls_crypt (`Msg "server only supports tls-crypt-v2")
+        | Ok (Packet.Hard_reset_client_v3, _key, received, linger) ->
+            assert (Cstruct.is_empty linger);
+            (* decode and unwrap wKc *)
+            let* actual_packet, wkc =
+              Tls_crypt.Wrapped_key.of_cstruct received
+            in
+            let* tls_crypt, _metadata =
+              Tls_crypt.Wrapped_key.unwrap ~key:wrapping_key wkc
+            in
```
hannesm, line 1957: you likely know better -- is the wkc encrypted as the other parts of the control packets? according to `handle_client`, it looks like it is encrypted in wrap_and_out calling encrypt_and_out

PR #247, opened 2024-05-15
Control channel explicit exit notifications

reynir:
- Parse better control channel messages

  Messages EXIT, RESTART and HALT are now better parsed. Messages RESTART and HALT can carry an optional message separated by a comma, e.g.

      HALT,shutting down

  The OpenVPN parser is very sloppy and only checks the prefixes "EXIT", "HALT", and "RESTART". Thus "HALTTTTT" is interpreted by OpenVPN as a HALT message. I chose not to be this sloppy with the parsing. I also chose not to attempt parsing the message for now. There are also potentially flags in the beginning of the message (only for RESTART according to the openvpn-rfc, but OpenVPN also parses it for HALT) - oddly they are not stripped from the message in OpenVPN. We can parse them at a later point if we so wish.

- Differentiate between `` `Exit `` actions triggered by e.g. timeouts and control channel messages.

The implementations are a bit too eager to exit. For `RESTART` (and maybe `EXIT`) we are expected to reconnect.

### DIFF
--- a/src/cc_message.ml
+++ b/src/cc_message.ml
@@ -1,0 +1,27 @@
+type cc_message = [ `Cc_restart | `Cc_halt | `Cc_exit ]
+
+(* OpenVPN is sloppy in its parsing and considers it a valid message if the
+   message just starts with {EXIT,HALT,RESTART}. We will not be as sloppy in
+   our parsing. *)
+let parse = function
+  | "EXIT\000" -> Some `Cc_exit
+  | "HALT\000" -> Some `Cc_halt
+  | "RESTART\000" -> Some `Cc_restart
+  | msg ->
+      if String.starts_with msg ~prefix:"RESTART," then
+        (* Ignoring message *)
+        Some `Cc_restart
+      else if String.starts_with msg ~prefix:"HALT," then
+        (* Ignoring message *)
+        Some `Cc_halt
+      else None
+
+let to_string = function
+  | `Cc_restart -> "RESTART\000"
+  | `Cc_halt -> "HALT\000"
+  | `Cc_exit -> "EXIT\000"
+
+let pp ppf = function
+  | `Cc_restart -> Fmt.string ppf "restart"
+  | `Cc_halt -> Fmt.string ppf "halt"
+  | `Cc_exit -> Fmt.string ppf "exit"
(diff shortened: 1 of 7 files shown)


src/cc_message.ml
```diff
@@ -0,0 +1,27 @@
+type cc_message = [ `Cc_restart | `Cc_halt | `Cc_exit ]
```
hannesm, line 1: I was curious whether restart and halt should carry a string option for the message?

src/engine.ml
```diff
@@ -825,35 +825,22 @@ let incoming_control_client config rng session channel now op data =
...
-              [] )
-      | None ->
-          (* a bit odd to return [None] ip_config *)
-          ( None,
-            config,
-            { channel with channel_st = Established (tls', keys) },
-            [] ))
+      let channel = { channel with channel_st = Established (tls', keys) } in
+      let act =
+        match d with
```
hannesm, line 833: maybe an Option.map looks nicer here?

src/engine.ml
```diff
@@ -825,35 +825,22 @@ let incoming_control_client config rng session channel now op data =
...
+        | None -> None
+        | Some d -> (
+            let d = Cstruct.to_string d in
+            match Cc_message.parse d with
```
hannesm, line 837: Here I can think of an Option.iter...

hannesm (approved):
looks good, minor remarks (but this can as well be merged as is, we can always improve later)

src/engine.ml
reynir: There is unfortunately no `Option.iter_none`, and it's in the `None`-case we want to log a warning :/

reynir:
I addressed the review comments. Unfortunately in https://github.com/robur-coop/miragevpn/pull/247/commits/6dd447af37dc500f86d912e90b9debf1c0c34854 the commit message is slightly wrong about `EXIT`. I'm not sure it's worth rewording before a merge...

hannesm:
looks good, feel free to merge (I think a squash-merge is fine)

reynir:
ocaml-ci is **very** slow today with the analysis step (almost half an hour by now)... I'll squash merge and then we can revert if needed.

PR #249, opened 2024-05-16
Miragevpn.server: authentication

hannesm:
This is used for password authentication.

### DIFF
--- a/src/config.ml
+++ b/src/config.ml
@@ -416,9 +416,14 @@ module Conf_map = struct
              if not (mem Ca t || mem Peer_fingerprint t) then
                Error
                  "server without a --ca or --peer-fingerprint, but \
-                  'verify-client-cert' is not set to 'required'"
+                  --verify-client-cert is set to 'required' (or not provided, \
+                  and required is the default)."
              else Ok ()
-         | _ -> Ok ()
+         | Some `Optional ->
+             Error
+               "server with --verify-client-cert optional, which is not \
+                supported at the moment. Please get in touch with the \
+                developers."
        in
        let* () =
          if mem Ca t && mem Peer_fingerprint t then
--- a/src/config.ml
+++ b/src/config.ml
@@ -404,6 +404,29 @@ module Conf_map = struct
          | None | Some (_, `Tcp _) -> Ok ()
          | Some (_, `Udp) -> Error "only TCP supported in the server"
        in
+       let* () =
+         match find Verify_client_cert t with
+         | Some `None ->
+             Log.warn (fun m ->
+                 m
+                   "server with 'verify-client-cert none', ensure a different \
+                    authentication mechanism is set up.");
+             Ok ()
+         | None | Some `Required ->
+             if not (mem Ca t || mem Peer_fingerprint t) then
+               Error
+                 "server without a --ca or --peer-fingerprint, but \
+                  'verify-client-cert' is not set to 'required'"
+             else Ok ()
+         | _ -> Ok ()
+       in
+       let* () =
+         if mem Ca t && mem Peer_fingerprint t then
+           Error
+             "While --ca and --peer-fingerprint are not mutually exclusive the \
+              semantics are unclear to us and thus not implemented"
+         else Ok ()
+       in
        Ok ())
 
   let is_valid_client_config t =
--- a/src/engine.ml
+++ b/src/engine.ml
@@ -318,6 +318,16 @@ let server server_config ~is_not_taken ?auth_user_pass server_ts server_now
     server_rng =
   let open Result.Syntax in
   let+ () = Config.is_valid_server_config server_config in
+  (match (auth_user_pass, Config.find Verify_client_cert server_config) with
+  | None, Some `None ->
+      Log.warn (fun m ->
+          m
+            "Server configuration without authentication! Your server accepts \
+             all clients. You should reconsider and use '--verify-client-cert \
+             required' and provide a '--ca' or '--peer-fingerprint'. \
+             Alternatively, provide ~auth_user_pass that checks for usernames \
+             and password.")
+  | _ -> ());
   let port = Option.value ~default:1194 (Config.find Port server_config) in
   ( {
       server_config;
--- a/src/engine.ml
+++ b/src/engine.ml
@@ -546,11 +546,26 @@ let kex_server config auth_user_pass session (my_key_material : my_key_material)
         Ok (auth ~user ~pass)
   in
   if not authenticated then
+    let td =
+      let options = Config.server_generate_connect_options config in
+      {
+        Packet.pre_master = Cstruct.empty;
+        random1 = my_key_material.random1;
+        random2 = my_key_material.random2;
+        options;
+        user_pass = None;
+        peer_info = None;
+      }
+    in
     let* tls', payload =
       Option.to_result ~none:(`Msg "not yet established")
-        (Tls.Engine.send_application_data tls [ Packet.auth_failed ])
+        (Tls.Engine.send_application_data tls [ Packet.encode_tls_data td ])
     in
-    Ok (`Authenticaion_failed tls', config, payload)
+    let* tls'', payload' =
+      Option.to_result ~none:(`Msg "not yet established")
+        (Tls.Engine.send_application_data tls' [ Packet.auth_failed ])
+    in
+    Ok (`Authentication_failed tls'', config, [payload; payload'])
 
   else
   let* cipher =
@@ -649,7 +664,7 @@ let kex_server config auth_user_pass session (my_key_material : my_key_material)
     | _ ->
         Error (`Msg "found Ifconfig without IPv4 addresses, not yet supported")
   in
-  (state, config, payload)
+  (state, config, [payload])
 
 let push_request tls =
   Option.to_result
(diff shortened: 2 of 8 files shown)


hannesm:
Two distinct commits, the first implementing username and password authentication (by using a callback passed to `Miragevpn.server`). The second installs a X509 authenticator to eventually request a client certifcate.

src/engine.ml
```diff
@@ -525,10 +533,18 @@ let tls_ekm tls config =
...
   let open Result.Syntax in
-  (* TODO verify username + password, respect incoming data *)
   let* their_tls_data = Packet.decode_tls_data ~with_premaster:true data in
+  let* () =
+    match (auth_user_pass, their_tls_data.user_pass) with
+    | None, _ -> Ok ()
+    | Some _, None -> Error (`Msg "no username and password provided")
+    | Some auth, Some (user, pass) ->
+        if auth ~user ~pass then Ok ()
+        else Error (`Msg "couldn't verify username and password")
```
reynir, line 546: Hmm. Should we would send `AUTH_FAILED` here? I'm not so sure about using `Error _` here.

reynir: The openvpn-rfc is confusing wrt. `AUTH_FAILED`, and the section on user/password authentication is [TBD](https://github.com/OpenVPN/openvpn-rfc/blob/master/openvpn-wire-protocol.xml#L1967-L1970)...

src/engine.ml
```diff
@@ -939,22 +939,48 @@ let server_handle_tls_data config auth_user_pass is_not_taken session keys tls d
...
+                m
+                  "server with 'verify-client-cert none', ensure a different \
+                   authentication mechanism is set up.");
+            Ok None
+        | Some `Optional, _ ->
+            Log.warn (fun m ->
+                m
+                  "server with 'verify-client-cert optional', ensure a \
+                   different authentication mechanism is set up.");
+            Ok (Some (fun ?ip:_ ~host:_ _certs -> Ok None))
```
reynir: We should probably track authentication status somehow and try verify the cert. If it's signed by the CA then the client doesn't need to authenticate using username and password. I don't know if ocaml-tls allows for this kind of maybe-authentication... (I can investigate)

src/engine.ml
```diff
@@ -951,31 +951,52 @@ let incoming_control_server auth_user_pass is_not_taken config rng session
...
+                           X509.Validation.trust_cert_fingerprint
+                             ~time:(fun () -> Some now)
+                             ~hash:`SHA256 ~fingerprint ?ip ~host chain)
+                     (Error (`Msg "No fingerprints provided"))
+                     fps))
+        | Some `None, _, _ ->
             Log.warn (fun m ->
                 m
                   "server with 'verify-client-cert none', ensure a different \
                    authentication mechanism is set up.");
```
reynir, line 982: We should log this warning in verify_server_config (or how the function is named). Otherwise it's logged on very new connection rather than just once.

src/engine.ml
```diff
@@ -951,31 +951,52 @@ let incoming_control_server auth_user_pass is_not_taken config rng session
...
             Ok None
-        | Some `Optional, _ ->
+        | Some `Optional, _, _ ->
             Log.warn (fun m ->
                 m
                   "server with 'verify-client-cert optional', ensure a \
                    different authentication mechanism is set up.");
             Ok (Some (fun ?ip:_ ~host:_ _certs -> Ok None))
-        | _, None ->
+        | _, None, None ->
```
reynir, line 990:
```suggestion
        | Some (`Optional | `Required), None, None ->
```

reynir (commented):
Thanks! I need to study what is expected to happen on authentication failure. For `--verify-client-cert optional` it should be sufficient to provide a valid certificate and we currently don't implement that.

src/engine.ml
hannesm: Oh, I was not aware of the semantics of openvpn with `verify-tls-cert optional`. Is the following the case:
- the server requests a certificate
- if the ca or the peer fingerprint matches, we're done (i.e. no further authentication)
- if it does not succeed, we do username and password authentication (and depending on its success/failure, we proceed or fail establishing the connection)

The only path I can think of right now is to have a `mutable authenticated : bool` in the type t, and set this from the authenticator callback. What do you think?

OCaml-TLS API is: you can provide an `authenticator`, which returns Ok or Error. On Ok, the TLS handshake is done further, on Error the handshake errors (and thus you don't get a TLS session).

src/engine.ml
hannesm: how should that be sent? I think we should send it to the client, but I was not aware how the encoding / on which channel this should be sent.

src/engine.ml
reynir: Yes, mostly.

* The server requests a certificate
* if the CA or peer-fingerprint matches the client is authenticated
* the client can send user-pass regardless of certificate authentication status and is (double) authenticated if the user-pass is valid

All this gets more complicated with all the callback scripts you can hook in into the OpenVPN authentication process. I don't know what you can do with the callback scripts and I'm afraid to look :D

> The only path I can think of right now is to have a mutable authenticated : bool in the type t, and set this from the authenticator callback. What do you think?

Hmm in that case we should be careful because we create new `t`s all the time.

Looking at the code the `authenticator` returns either `Ok None` or `Ok (Some leaf_cert_or_is_it_cert_chain)`. If so we can maybe check after the handshake if the tls object has a peer certificate and in that case consider the peer authenticated.

src/engine.ml
hannesm: done in 1a851ea

src/engine.ml
reynir: Yes, it's a big footgun. We could as well in `Miragevpn.server` fail with an error on `verify-tls-cert optional` and `auth_user_pass = None`.

I think the `verify-tls-cert optional` could be useful if you have some (e.g. headless) clients where a certificate is more suitable for authentication mixed with clients with a human user behind that can input username and password (but can't remember their certificate and key).

So it's not clear to me if we want to support `verify-tls-cert optional` or not.

src/engine.ml
hannesm: thanks for your investigations and fixes, this is done in later commits ([d5e1be4](https://github.com/robur-coop/miragevpn/pull/249/commits/d5e1be42e762c85a1dd91b2ede1df7d37329facb) [37e4832](https://github.com/robur-coop/miragevpn/pull/249/commits/37e483239eff6e394b04e78b971509c58ad5647a))

hannesm:
from my side, good to squash & merge

reynir:
Thanks! The changes look good to me as reviewed from my phone. I can take a
closer look tomorrow, but otherwise this looks good to (squash) merge to me.

tor. 16. maj 2024 20.08 skrev Hannes Mehnert ***@***.***>:

src/config.ml
```diff
@@ -404,6 +404,34 @@ module Conf_map = struct
...
+               "server with --verify-client-cert optional, which is not \
+                supported at the moment. Please get in touch with the \
+                developers."
+       in
+       let* () =
+         if mem Ca t && mem Peer_fingerprint t then
+           Error
+             "While --ca and --peer-fingerprint are not mutually exclusive the \
+              semantics are unclear to us and thus not implemented"
+         else Ok ()
```
reynir, line 433: Good! I forgot about that one.

src/engine.ml
```diff
@@ -906,36 +958,72 @@ let server_send_push_reply config is_not_taken tls session key tls_data =
...
       let* ip_config, config, channel_st, out' =
         server_send_push_reply config is_not_taken tls' session key tls_data
       in
-      Ok (ip_config, config, channel_st, (`Control, out) :: out')
+      Ok (ip_config, config, channel_st, out @ out')
   | `State (channel_st, ip_config) ->
       (* keys established, move forward to "expect push request (reply with push reply)" *)
-      Ok (ip_config, config, channel_st, [ (`Control, out) ])
+      Ok (ip_config, config, channel_st, out)
+  | `Authentication_failed _tls -> Ok (Some `Exit, config, Expect_reset, out)
```
reynir, line 977: Very nice with the `` `Exit `` here. It didn't require as big a change as I thought \o/

I put `Expect_reset` to avoid sort of reset the state. I'm not sure it's sound to reuse the state on a reconnect, and we emit an `` `Exit `` now so the caller shouldn't reuse the state.

src/engine.ml
```diff
@@ -906,36 +958,72 @@ let server_send_push_reply config is_not_taken tls session key tls_data =
...
-        ( Config.get Ca config,
-          Config.get Tls_cert config,
-          Config.get Tls_key config )
+      let server, key =
+        (Config.get Tls_cert config, Config.get Tls_key config)
       in
       let ciphers = tls_ciphers config and version = tls_version config in
+      let* authenticator =
+        match
+          ( Config.find Verify_client_cert config,
```
reynir, line 992: Can be done in a different commit, but `verify-client-cert required` is the default according to the man page.

reynir: It does not make a big difference whether we synthesize here or in config.ml...

src/engine.ml
hannesm: I agree. Alternatives include:
- allow the error case to carry an out packet
- allow the ok case to have no state (and no config?)

I'd go for the first one if asked, but don't see any advantage right now to refactor.

src/engine.ml
hannesm: yes, well, there's the pain of matching on ``None | Some `Required`` -- but I guess a refactoring may be to move from config to some custom record where only the stuff really needed is present... at the cost of having yet another record type

or to fill the verify-client-cert in valid_server_config if not provided, and use here a `Config.get` instead of `Config.find`.

PR #250, opened 2024-05-16
On user-pass auth failure send AUTH_FAILED

reynir:
This is not the right way to do it, and we should emit an `Exit action.

### DIFF
--- a/src/engine.ml
+++ b/src/engine.ml
@@ -546,11 +546,26 @@ let kex_server config auth_user_pass session (my_key_material : my_key_material)
         Ok (auth ~user ~pass)
   in
   if not authenticated then
+    let td =
+      let options = Config.server_generate_connect_options config in
+      {
+        Packet.pre_master = Cstruct.empty;
+        random1 = my_key_material.random1;
+        random2 = my_key_material.random2;
+        options;
+        user_pass = None;
+        peer_info = None;
+      }
+    in
     let* tls', payload =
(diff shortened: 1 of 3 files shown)


reynir:
Experimenting a bit I learned the server should send its key exchange message **before** sending `` `AUTH_FAILED ``.

src/engine.ml
```diff
@@ -537,14 +537,37 @@ let kex_server config auth_user_pass session (my_key_material : my_key_material)
     tls data =
   let open Result.Syntax in
   let* their_tls_data = Packet.decode_tls_data ~with_premaster:true data in
-  let* () =
+  let* authenticated =
     match (auth_user_pass, their_tls_data.user_pass) with
-    | None, _ -> Ok ()
-    | Some _, None -> Error (`Msg "no username and password provided")
+    | None, _ -> Ok true (* erm... *)
```
hannesm, line 542: this case needs to be there as it is, since there could be servers that only use `verify-client-cert` -- so the server must be fine with no auth-user-pass being provided.

we can of course when creating the server check and warn if verify-client-cert is 'none' and auth-user-pass is not provided (and print a big warning that no authentication will take place).

hannesm: done in 238c2e9dea4ec7d20096d591e1865da261ced55d

hannesm:
thanks, I pushed this to the other PR. And added two minor commits.

PR #254, opened 2024-05-23
Push directive

reynir:
This implements the `--push` configuration directive.

I think maybe our `a_param` is not correct, and I will take a look at that.

### DIFF
--- a/src/config.ml
+++ b/src/config.ml
@@ -475,10 +475,13 @@ module Conf_map = struct
          "server tls-crypt-v2 key passed in tls-crypt-v2")
 
   let pp_param ppf s =
-    (* XXX(reynir): a_param backslash escaping is probably not correct *)
+    (* XXX(reynir): a_single_param backslash escaping is not correct. Spaces
+       (as considered by isspace()) are not correctly unescaped, and all other
+       escape sequences should be illegal but are accepted. Thus we only escape
+       double quotes here. *)
     let escape ppf s =
       Fmt.char ppf '"';
-      Fmt.(list ~sep:(any "\\\"") string) ppf (String.split_on_char '"' s);
+      Fmt.(list ~sep:(any {|\"|}) string) ppf (String.split_on_char '"' s);
       Fmt.char ppf '"'
     in
     if String.exists (function '"' | '\'' | ' ' -> true | _ -> false) s then
--- a/src/config.ml
+++ b/src/config.ml
@@ -212,6 +212,7 @@ module Conf_map = struct
     | Pkcs12 : X509.PKCS12.t k
     | Port : int k
     | Pull : flag k
+    | Push : string list k
     | Proto
         : ([ `Ipv6 | `Ipv4 ] option
           * [ `Udp | `Tcp of [ `Server | `Client ] option ])
@@ -473,6 +474,17 @@ module Conf_map = struct
        ensure_not Tls_crypt_v2_server
          "server tls-crypt-v2 key passed in tls-crypt-v2")
 
+  let pp_param ppf s =
+    (* XXX(reynir): a_param backslash escaping is probably not correct *)
+    let escape ppf s =
+      Fmt.char ppf '"';
+      Fmt.(list ~sep:(any "\\\"") string) ppf (String.split_on_char '"' s);
+      Fmt.char ppf '"'
+    in
+    if String.exists (function '"' | '\'' | ' ' -> true | _ -> false) s then
+      escape ppf s
+    else Fmt.string ppf s
+
   let pp_opt_direction ppf = function
     | None -> ()
     | Some `Incoming -> Fmt.pf ppf " 1"
@@ -649,6 +661,10 @@ module Conf_map = struct
               (append (any " ") (of_to_string Protocol_flag.to_string)))
           flags
     | Pull, () -> p () "pull"
+    | Push, push_options ->
+        p () "%a"
+          Fmt.(list ~sep (fun ppf -> pf ppf "push %a" pp_param))
+          push_options
     | Remote, lst ->
         p () "%a"
           Fmt.(
@@ -1642,6 +1658,23 @@ let a_remote :
 
 let a_remote_entry = a_remote
 
+let a_push =
+  string "push" *> a_whitespace *> commit *> a_single_param
+  >>= fun push_option ->
+  (* the OpenVPN man page says only a subset of options can be pushed.
+     This is however misleading as the server can push what it pleases.
+     Only, the client will filter out unaccepted options.
+
+     The only constraint seems to be the option can't contain commas as
+     comma will be used as a separator when pushing options. *)
+  if String.contains push_option ',' then
+    let msg =
+      Fmt.str "Push option contains illegal comma (',') in string: %S"
+        push_option
+    in
+    return (`Ignored msg)
+  else return (`Entry (B (Push, [ push_option ])))
+
 let a_connection =
   skip_many (a_whitespace_or_comment <|> a_newline) *> a_remote_entry
   <* skip_many (a_whitespace_or_comment <|> a_newline)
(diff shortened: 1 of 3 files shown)


src/engine.ml
```diff
@@ -938,10 +939,12 @@ let server_send_push_reply config is_not_taken tls session key tls_data =
...
-             Fmt.str "%a" (Config.pp_b ?sep:None)
-               (Config.Conf_map.B (Protocol_flags, flags))))
+    @ (Config.find Protocol_flags config
+      |> Option.map (fun flags ->
+             Fmt.str "%a" (Config.pp_b ~sep)
+               (Config.Conf_map.B (Protocol_flags, flags)))
+      |> Option.to_list)
+    @ Option.value ~default:[] (Config.find Push config)
   in
   let reply = String.concat "," reply_things in
```
reynir, line 949: We may consider creating a `Config.Conf_map.t` instead...

src/engine.ml
```diff
@@ -938,10 +939,12 @@ let server_send_push_reply config is_not_taken tls session key tls_data =
...
-    @ (Option.to_list (Config.find Protocol_flags config)
-      |> List.map (fun flags ->
-             Fmt.str "%a" (Config.pp_b ?sep:None)
-               (Config.Conf_map.B (Protocol_flags, flags))))
+    @ (Config.find Protocol_flags config
+      |> Option.map (fun flags ->
+             Fmt.str "%a" (Config.pp_b ~sep)
+               (Config.Conf_map.B (Protocol_flags, flags)))
+      |> Option.to_list)
+    @ Option.value ~default:[] (Config.find Push config)
```
reynir, line 947: It's not so nice that this will create e.g. `"push foo,push bar"` if we have `Push, ["foo", "bar"]`

src/config.ml
```diff
@@ -473,6 +474,17 @@ module Conf_map = struct
...
+    (* XXX(reynir): a_param backslash escaping is probably not correct *)
+    let escape ppf s =
+      Fmt.char ppf '"';
+      Fmt.(list ~sep:(any "\\\"") string) ppf (String.split_on_char '"' s);
+      Fmt.char ppf '"'
+    in
+    if String.exists (function '"' | '\'' | ' ' -> true | _ -> false) s then
+      escape ppf s
+    else Fmt.string ppf s
+
```
reynir, line 487: here I don't escape backslashes as `a_param` seemingly doesn't "unescape" backslashes except for the quote character.

src/engine.ml
hannesm: does this matter? should we revise in here (should be easy)?

reynir: No, it doesn't matter much.

hannesm: hmm, actually I don't quite understand. `Push` is a string list. What does the other side expect - is it the string of the values "foo, bar" or "push foo,push bar"?

As I read the code here, if we have "push foo" "push bar" in the configuration (which will turn into `Push ["foo", "bar"]`, we will transmit "foo, bar" over the wire.

src/config.ml
reynir: To be more precise: when the parser in a quoted string encounters `'\\' '"'` it produces `"\""`. For any other character `c`, `'\\' c` produces `"\\c"`, that is the backslash is kept. So `"\\ "` is parsed as `"\\ "` and not `" "`. That is why I only escape quote characters above.
https://github.com/robur-coop/miragevpn/blob/807ab098c72887939ac036917451f2bd0c9e1c7e/src/config.ml#L886-L891

src/config.ml
```diff
@@ -1642,6 +1658,23 @@ let a_remote :
...
+     Only, the client will filter out unaccepted options.
+
+     The only constraint seems to be the option can't contain commas as
+     comma will be used as a separator when pushing options. *)
+  if String.contains push_option ',' then
+    let msg =
+      Fmt.str "Push option contains illegal comma (',') in string: %S"
+        push_option
+    in
+    return (`Ignored msg)
```
hannesm, line 1675: ignored or error? I'm a fan of fail early and fail hard.

src/engine.ml
reynir: a `push foo` and `push bar` tells OpenVPN to push `foo` and `bar`. In other words the push reply will look like `PUSH_REPLY,..,foo,bar,..\000` (or any permutation thereof; I don't think the order is important)

src/config.ml
hannesm: If `\"` is in a string, this will be converted to `\\"`. I'm fine until we hit an issue with that.

src/config.ml
```diff
@@ -473,6 +474,17 @@ module Conf_map = struct
        ensure_not Tls_crypt_v2_server
          "server tls-crypt-v2 key passed in tls-crypt-v2")
 
+  let pp_param ppf s =
+    (* XXX(reynir): a_param backslash escaping is probably not correct *)
+    let escape ppf s =
+      Fmt.char ppf '"';
+      Fmt.(list ~sep:(any "\\\"") string) ppf (String.split_on_char '"' s);
```
hannesm, line 481: I find it more readable to use the other string syntax:

```suggestion
      Fmt.(list ~sep:(any {|\"|}) string) ppf (String.split_on_char '"' s);
```

src/engine.ml
hannesm, line 949:
```suggestion
  let reply = String.concat sep reply_things in
```

src/config.ml
reynir: As I recall I found that this is the behavior of OpenVPN. We may reconsider this...

src/engine.ml
reynir: `sep` is not in scope here anymore and is a `unit Fmt.t` and not a `string`.

src/engine.ml
```diff
@@ -938,10 +939,12 @@ let server_send_push_reply config is_not_taken tls session key tls_data =
     (* XXX(reynir): here we assume the binding for [Key_derivation] is only ever [`Tls_ekm] *)
     @ (if Config.mem Key_derivation config then [ "key-derivation tls-ekm" ]
        else [])
-    @ (Option.to_list (Config.find Protocol_flags config)
-      |> List.map (fun flags ->
-             Fmt.str "%a" (Config.pp_b ?sep:None)
-               (Config.Conf_map.B (Protocol_flags, flags))))
+    @ (Config.find Protocol_flags config
+      |> Option.map (fun flags ->
+             Fmt.str "%a" (Config.pp_b ~sep)
```
hannesm, line 944: Why did the `sep` change here? In the end, it doesn't matter, since we'll output "protocol-flags%a", and using `Fmt.(list ~sep:nop (append (any " ") (of_to_string Protocol_flag.to_string))` applied to `flags`. So there won't be ever a `~sep` being printed.

src/engine.ml
hannesm: ah, ok

src/engine.ml
reynir: The `sep` was wrong as it would have output a newline instead of a comma between the configuration directives. Thankfully, it would only ever output one configuration directive. So changing `sep` here does not actually change anything, but it is the morally correct code as we should, in this context, use commas to separate configuration directives.

src/config.ml
reynir: Addressed in f2b331bfd68139337f4060f87503c2bf3c900ace

src/config.ml
reynir: I expanded on the comment in f2b331bfd68139337f4060f87503c2bf3c900ace. I would revisit this when addressing #257.

src/engine.ml
reynir: With this in mind maybe we should not have `~sep` be optional...

src/config.ml
hannesm: addressed in https://github.com/robur-coop/miragevpn/commit/cd7d999321e13993862af649977689aa96a7e114

PR #264, opened 2024-06-17
Improve openvpn_config_parser

reynir:
We can now validate both client and server configurations, and the block size padding is now configurable.

### DIFF
--- a/app/openvpn_config_parser.ml
+++ b/app/openvpn_config_parser.ml
@@ -21,19 +21,21 @@ let read_config_file fn =
   | Ok str -> parse_client ~string_of_file str
   | Error _ as e -> e
 
-let alignment_header =
-  {|#############
+let alignment_header block_size =
+  Printf.sprintf
+    {|#############
 # IMPORTANT:
 #  This OpenVPN configuration file has been padded with the comment below to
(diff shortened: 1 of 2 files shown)


app/openvpn_config_parser.ml
```diff
@@ -46,44 +48,90 @@ let pad_output output =
...
+      if v <= 0 then Error (`Msg "Non-positive integer") else Ok v
+    and print = Arg.(conv_printer int) in
+    Arg.conv (parse, print) ~docv:"POSINT"
+  in
+  let doc = "The block size to align against" in
+  Arg.(value & opt pos_int 512 & info [ "block-size" ] ~doc ~docv:"BLOCKSIZE")
+
+let cmd =
+  let term = Term.(const jump $ setup_log $ config $ mode $ block_size)
+  and info = Cmd.info "openvpn_config_parser" ~version:"%%VERSION_NUM" in
```
hannesm, line 134:
```suggestion
  and info = Cmd.info "openvpn-config-parser" ~version:"%%VERSION_NUM" in
```

the public name uses `-`, not `_`.

hannesm:
looks fine to me. thanks a lot.

PR #266, opened 2024-06-18
Some enhancements for the miragevpn_client_lwt

hannesm:
There are 3 things:
- on FreeBSD, chop the 4 bytes header from tun_read, and add 4 bytes header on write
- add a host route to the VPN server using the old default gateway (this part has not been tested on Linux, please go ahead and do so)
- catch Ctrl-C and run at_exit on Ctrl-C

Hope this is in your interest :)

### DIFF
--- a/app/miragevpn_client_lwt.ml
+++ b/app/miragevpn_client_lwt.ml
@@ -292,14 +292,31 @@ let send_recv conn config ip_config _mtu routes =
         (* not using write_to_fd here because partial writes to a tun
            interface are semantically different from single write()s: *)
         Lwt_list.iter_p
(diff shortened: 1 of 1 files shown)


hannesm:
There's still an issue.. the miragevpn_client_lwt doesn't do any IP address rewriting, and our server complains about "[WARNING] [miragevpn.mirage] my-broadband-ip:57535 received payload where source ip 192.168.42.80 (the IP of my wlan device) doesn't match expected 10.8.0.2 (the IP of my tun device)".

hannesm:
The last commit fails hard if something unexpected is put into est_mvar

reynir:
Thanks!

PR #268, opened 2024-06-20
Compute routes with net_gateway or remote_host

reynir:
And attempt autolocal.

Instead of the `` `Established _ `` action returning a list of routes it returns an opaque object of type `route_info` which under the hood is `Config.t`. Then the client of the engine can call `Miragevpn.routes` passing the `route_info` as well as some required information on the remote host ip and routing table. This way we can resolve magic keywords `net_gateway` and `remote_host` in `--route` directives, and we attempt to figure out `autolocal`. We also respect the `local` flag in `--redirect-gateway` if present.

I learnt that OpenVPN discards routes which it can't resolve, for example a `--route` with `net_gateway` on an IPv6-only host. Thus I decided to warn and skip such routes. The `net_gateway` and `remote_host` are `Ipaddr.V4.t option`s in case of IPv6.

Certainly the route detection in miragevpn-client-lwt could be made more robust.

### DIFF
--- a/app/miragevpn_client_lwt.ml
+++ b/app/miragevpn_client_lwt.ml
@@ -2,6 +2,10 @@ open Lwt.Infix
 
 let error_msgf fmt = Fmt.kstr (fun msg -> Error (`Msg msg)) fmt
 
+let pp_route ppf (net, gw, metric) =
+  Fmt.pf ppf "%a via %a metric %u" Ipaddr.V4.Prefix.pp net Ipaddr.V4.pp gw
+    metric
+
 type supported = FreeBSD | Linux
 
 let platform =
@@ -15,7 +19,7 @@ let platform =
 
 let connected_ip = ref Ipaddr.V4.any
 
-let remote_host_route dst =
+let shares_subnet dst =
   let dst = Ipaddr.V4.to_string dst in
   match Lazy.force platform with
   | Linux -> (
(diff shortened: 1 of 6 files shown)


reynir:
While testing this (before I had removed `remote_host_route`) I discovered on Linux at least if you try to add the exact same route twice `ip route` exits with code 2 and a message `RTNETLINK answer: file exists`. I found there is as well `ip route replace` which may be worth looking into.

hannesm (approved):
looks good, didn't test it, though

app/miragevpn_client_lwt.ml
```diff
@@ -15,29 +19,43 @@ let platform =
...
+          match String.split_on_char ' ' ip with
+          | "default" :: "via" :: ip :: _ -> Some ip
+          | _ -> None)
+      | Error (`Msg m) -> invalid_arg ("couldn't find default route " ^ m))
+  | FreeBSD -> (
+      let cmd = Bos.Cmd.(v "route" % "-n" % "show" % "default") in
+      match Bos.OS.Cmd.(run_out cmd |> out_lines |> success) with
+      | Ok (_ :: _ :: _ :: gw :: _) ->
+          Option.map String.trim (List.nth_opt (String.split_on_char ':' gw) 1)
+      | Ok _ -> None
```
hannesm, line 58:
```suggestion
    | Ok lines -> (
          match
            List.find_opt
              (fun l -> String.starts_with ~prefix:"gateway:" (String.trim l))
              lines
          with
          | Some gw ->
              Some (String.trim (List.nth (String.split_on_char ':' gw) 1))
          | None -> None)
```

I find the String.nth and `_::_::_::gw::_` a bit too handwavy (and a bit too tied to the current output of route).

hannesm:
great!

PR #269, opened 2024-06-21
Bump cmdliner version constraint

reynir:
Due to usage of Cmdliner.Term.Syntax introduced in 4d96f2bdebfe53ff6b97b152292403c14adc5dc7 which is new in cmdliner.1.3.0.


reynir:
Alternatively we can drop the let-syntax...

hannesm:
looks good to me, thanks

PR #270, opened 2024-06-24
end-to-end testing shell scripts

hannesm:


hannesm:
The last commit re-adds AES-GCM to tls-crypt in data-ciphers, since the GitHub runner is ubuntu 22.04 with openvpn 2.5.9 which doesn't support poly1305-chacha20. I tried to use ubuntu 24.04, but that failed to install darcs (which somehow opam requires in the setup-ocaml CI action).

Anyways, coverage around 72%.

reynir:
I think this is very nice now!

PR #272, opened 2024-06-25
Add and use pkcs12 in tls-crypt-v2 e2e test

reynir:
This will exercise some of the pkcs12 code.

I also removed unused symbolic links.


reynir:
I added a test with `peer-fingerprint` too, but `--peer-fingerprint` is OpenVPN>=2.6 so doesn't work with CI.

hannesm:
I opened #273 to get openvpn 2.6 installed on the GitHub worker (following the OpenVPN instructions). I guess that would be nice for this here as well :)

test/e2e/miragevpn-server.sh
```diff
@@ -28,9 +28,9 @@ trap cleanup EXIT
 
 (
 	sleep 1
-	openvpn --cd "$config_dir" --config "client.conf" --dev-type tun --dev "$tun_interface" --writepid "$pidfile" --script-security 2 --up ../client-up.sh > /dev/null
+	openvpn --cd "$config_dir" --config "client.conf" --dev-type tun --dev "$tun_interface" --writepid "$pidfile" --script-security 2 --up ../client-up.sh
```
hannesm, line 31: I'm undecided about the output here.. somehow it is nice to spot the issues easily in CI, somehow it is also nice to not get a lot of output.

In the end, I'm fine without the redirection to /dev/null. if there's need to see less output, someone invoking the scripts can still redirect to /dev/null. :)

hannesm:
looks good to me

test/e2e/miragevpn-server.sh
reynir: I removed the redirection to /dev/null to debug in CI. We should add it back, I just forgot.

reynir:
The two numbers are unfortunately not 100% comparable as for the 2nd numbers I rebased on #273 which covers more code as well. While rebasing I also dropped the "debug" commit which removed the redirection to /dev/null.

hannesm:
The other PR was at 72.76%, here we're nearly 1% more.

reynir:
I squashed the fixup commit before merging

hannesm:
Should we remove the redirects? As mentioned above, fine with me.

PR #273, opened 2024-06-25
end-to-end: use openvpn repo to retrieve a more recent version

hannesm:
using the howto from https://community.openvpn.net/openvpn/wiki/OpenvpnSoftwareRepos#DebianUbuntu:UsingOpenVPNaptrepositories


reynir (approved):
Thanks! This is great! \o/

PR #223, opened 2024-04-04
Don't use cstruct - use string/bytes!

reynir:

PR #280, opened 2024-09-27
Various optimizations

reynir:
These are mostly optimizations from https://github.com/robur-coop/miragevpn/pull/223 but also some new ones since we have encrypt_into etc. functions.

`main-bench.exe` is built from 6099a4bd4134eedddd16090a083dc58d246eeae1 (unfortunately without --release)
image

As you can see above there are improvements in execution time, but most notably the number of minor words allocations is almost halved. When testing against an openvpn server I didn't observe much of a difference except, curiously, a slowdown in upload speed (from 826 Mb/s to 757 Mb/s). Why that would I don't know.

### DIFF
--- a/bench/bench_engine.ml
+++ b/bench/bench_engine.ml
@@ -194,7 +194,7 @@ let test_receive_data cipher =
     Staged.stage @@ fun () ->
     match Miragevpn.handle established_client (`Data pkt) with
     | Ok _ -> ()
-    | Error _ -> assert false
+    | Error err -> Format.kasprintf failwith "%a" Miragevpn.pp_error err
   in
   Test.make ~name:"decode data" staged
 
--- a/src/engine.ml
+++ b/src/engine.ml
@@ -2246,7 +2246,12 @@ let handle_static_client t s keys ev =
               | Error `Tcp_partial ->
                   (* we don't need to check protocol as [`Tcp_partial] is only ever returned for tcp *)
                   Ok ({ t with linger }, acc)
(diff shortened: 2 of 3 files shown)


reynir:
> When testing against an openvpn server I didn't observe much of a difference except, curiously, a slowdown in upload speed (from 826 Mb/s to 757 Mb/s). Why that would I don't know.

Ah! This was with the UDP configuration. With TCP there is a slight speedup (down: 919 Mb/s -> 1.07 Gb/s, up: 599 Mb/s -> 652 Mb/s)

reynir:
I accidentally committed and pushed 15ba1eda481228c8560eacd8b7a12e825d271af2 and eba6e587fac9b1756b63e92e13e327f546208e24 to `main`. They were a fix to the failing e2e test where I observed locally that it failed due to a Data_v1 packet with key id 0 arriving seemingly just after a successful rekey for key id 1. I don't know if it's the correct thing to do, but to me it seems for data packets we could be more lenient and just drop them.

reynir:
Ah wait the e2e test already fails at the static client configuration. Locally, it initially also failed there for me, but that was due to firewall rules.

hannesm:
looks fine to me, may revert the e2e changes and redirect again to /dev/null

reynir:
I cleaned up the git history a bit by merging two commits and dropping the CI debugging commit.

It fails on OCaml 5.3~alpha1 due to `effect` suddenly being a keyword. I'll address this separately.

PR #291, opened 2026-05-23
Tls plain

hannesm:
should help #36 #288

testing with udp and openvpn, requres data v2 which I'll do soon.


hannesm:
This now seems to work fine in the p2p udp mode with tls-plain. I haven't tested in other scenarios, and am not 100% sure about when the peer_id is set where and how...

NCP_P2P is also required (a bit in IV_PROTO) to get EKM and IV_CIPHER there, ignoring the server configuration sent by the server...

happy to hear your thoughts. I may rewrap my head later to write a more explanatory comment here, on what and why.

hannesm:
To start the story, I got a message from our VPN provider that they want to update our OpenVPN configuration. Previously, it was all static keys (no TLS), UDP, AES-256-CBC [in a p2p mode].

Now, it is a slightly different: using TLS, still p2p, still UDP, and no tls_auth/tls_crypt. I noticed that both #36 and #288 have this feature request as well.

So I started myself on the `tls-office.conf` / `tls-home.conf` from OpenVPN, found Reynir's branch, and figured out what to do. It was clear that quite some work is needed, since the no-tls-auth work does slightly different control packets (no HMAC) - see Packet.Tls_plain. It turned out that there's quite some hardcoded offset computations going on in that module, and there were various late-night issues with wrong offsets and length computations. But eventually I got this through (using tls-office.conf from this repository with the last line removed (no `proto tcp-server`) running on my laptop with OpenVPN 2.6.14.

On the other side, I found the app/miragevpn_client_notun.exe very appealing: takes a configuration file (tls-home-no-auth.conf), doesn't mess around with tun devices, and just sends ICMP ping (and waits for replies) over the tunnel to the gateway / other endpoint.

Nice, I thought. But OpenVPN complained `Data Channel Offload doesn't support DATA_V1 packets. Upgrade your server to 2.4.5 or newer.`. So I thought challenge accepted.

Now, DATA_V2 packets are the new great thing in OpenVPN. As I understand, the main advantage is the introduction of a `peer_id` to have a floating client / roadwarrior setup. It turns out AEAD is applied slightly differently -- and not too obvious what is done there, but the `additional data (adata)` includes not only the replay_id, but as well the key_op byte and the peer_id. I figured that out by the same setup as above, and sending from my host machine a ping to 10.2.0.2 -- and then looking at the hexdump and trying out what is going on (until the decrypt_verify succeeded).

It also seemed that by default, TCP is used, and some code needed to be there (in Engine.client) to decide whether TCP or UDP should be used.

I didn't look at the CBC and peer_id. Also, it was a bit non-obvious where to take the peer_id from, I figured it should default to "vpn" and "\xff\xff\xff" if unknown -- but also the server should hand it out in a PUSH_REPLY (hint: I don't get any PUSH_REPLY).

I also figured I should somehow get the cipher, so I finished the "merge_server_config" (from tls_data.options), with not 100% luck since the server unless manually configured would tell me BF-CBC :/ [nevertheless, it lead me to implement Key_method, Keysize, Peer_id in the config.ml, add support for uppercase "proto", etc.]

I also figured that in p2p mode (still not clear whether this is the default with the tls-office conf or what is happening there), the `ifconfig` stanza refers to local IP and peer IP, not a IP/netmask -- so some fixes are done in config_ext.ml for this.

Then I learned a bit more (by reading C code) about the IV_PROTO_NCP_P2P - and IV_PROTO_PEER_ID (also known as IV_PROTO_DATA_V2), and found that it seems the "merge_server_config" isn't really done, but instead the tls_data.peer_info should be properly handled (for IV_PROTO and IV_CIPHERS)... so, with that I get EKM and cipher - also if Peer_id is there as bit, it'll be `vpn`, and if EKM is enabled, it will be from the exporter secret. Now, some question remains about whether we always want to keep the peer_id mode around, or whether DATA_V1 packets are useful eventually... time will tell ;)

I'm very happy that I'm able to establish the tunnel, and would love to put this into production (and move our old VPN tunnel to the new one). Give me some more hours ;)

Looking at the diff, I'm curious:
- [x] key-method has been removed in OpenVPN 2.5 for security reasons and shouldn't be used -- maybe drop and ignore again? -- removed in 3e2d26b
- [x] keysize: ? undocumented -- removed in 3e2d26b
- [ ] is there a setup where peer_id is actually passed from server to client?
- [x] should peer_id be a `string option`? -- done in 3034bd8
- [ ] support for `[null-digest]` in auth (as in tls_plain) -- but again this is in the tls_data.options which doesn't seem to be something that anyone uses in such a setting -- so maybe we don't need to parse this?
- [x] what about replay/retransmissions? --- looks fine
- [x] what about rekeying intervals (or should we be lazy and wait for the other peer to act)? -- looks fine (done in timer etc)
- [x] there's no replay_id for control packets now, so should that be an int32 option? we have an ...`if cc = Tls then Ok () else`.. for now :) -- fine
- [ ] is there a smart and efficient way to condense Packet encode/decode of control (and header) independent of tls_crypt/tls_auth/tls being used?
- [ ] CBC and peer_id
- [x] manually tested with tests/e2e
- [ ] there's still some peer_id confusion... when using the tls-crypt examples in test/e2e, they break if the IV_PROTO includes Peer_id. unclear why - the server replies with peer-id 0 (which looks fishy), and according to the spec https://openvpn.github.io/openvpn-rfc/openvpn-wire-protocol.html#section-6.8 the peer should say in their peer_info `ID=<hex>`...

I'll keep on thinking about some of these issues.

palainp:
Dear @hannesm, thank you for your huge work on this PR. I tried running a test with qubes-miragevpn, but it didn't work :(

This library depends on bisect_ppx, and I encountered issues with cmdliner>=2.0.0 and bisect_ppx (there’s a 1yo pull request pending on bisect_ppx, but it hasn’t been merged; the current main branch seems to support cmdliner 2.0.0, at least according to the latest commit), and the recent mirage tool requires an up-to-date version of cmdliner.
But when I tried to pin the bisect_ppx main branch, the build failed :(

So for the time being I’m going to try removing this dependency for testing purposes, and I’ll get back to you with updates :)

hannesm:
Thanks @palainp. I added a commit here to remove the bisect_ppx dependency.

In the configuration you talk about in #288, there's still the issue of AES-128-CBC which is not supported with this PR. So best if you've a different configuration with AEAD ciphers ;)

Also, as described earlier, in our setup it is a p2p connection, not topology subnet. I'm not sure whether this makes a difference though...

palainp:
> Thanks @palainp. I added a commit here to remove the bisect_ppx dependency.
> 

Thank you!

> In the configuration you talk about in #288, there's still the issue of AES-128-CBC which is not supported with this PR. So best if you've a different configuration with AEAD ciphers ;)
> 

I've either managed to find a configuration that works, and/or added some code to the configuration parser to handle this (actually, I think https://github.com/palainp/miragevpn/tree/aes_128_cbc was my current status, but I need to see what is working or not there)

> Also, as described earlier, in our setup it is a p2p connection, not topology subnet. I'm not sure whether this makes a difference though...
>

I'll tell you where I can go :)

hannesm:
note, this is now deployed as robur:as250 with:
```
tls-client
dev    tun
proto  udp
rport  65222
remote AAAAAAA
ping   10
ping-restart 30
tun-mtu 1500
float

<cert>
</cert>

<key>
</key>

peer-fingerprint  ABCDE
#tls-cert-profile  preferred <- Hannes: I commented this out, since I've no clue what to do there

ifconfig  A B #pretty useless I think (these IP addresses don't have much value AFAICT)

verb 3
```

hannesm:
the last two commits are for rekeying, and the very last one is indeed an oversight that has been sitting in the codebase for quite some time undetected.... if there's a rekey, the idea is that we keep the old keys around for some time (in lame_duck) for packets from the other peer that are still encrypted with the old key material to be read and forwarded...

somehow we first updated the t.channel, and only thereafter instantiated t.lame_duck with t.channel -- so it was always the new channel in there.

